PR #8, opened 2014-06-03
[VFS-467] Improved the comparison of file names

coder-hugo:
- remove the trailing slash of the file names before comparing them

### DIFF
--- a/commons-vfs2/pom.xml
+++ b/commons-vfs2/pom.xml
@@ -98,7 +98,6 @@
     <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
-      <scope>test</scope>
     </dependency>
     <!-- Test FTP with Apache FTP Server (MINA) -->
     <dependency>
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/commons/vfs2/provider/webdav/WebdavFileObject.java
```diff
@@ -799,4 +795,25 @@ private String toUrlString(final URLFileName name, final boolean includeUserInfo
             return name.getURI();
         }
     }
+
+
+    private String removeTrailingSlash(final String fileName)
+    {
+        return StringUtils.removeEnd(fileName, "/");
```
kinow: I like the idea of simplifying the method and adding other simpler methods.

However, when I went to confirm if Commons Lang was a dependency, it appears it is, but a comment suggests it to be a test dependency? https://github.com/apache/commons-vfs/blob/0dd96fde5eebbf14fedef6c7aa110979d7efe14b/pom.xml#L484

Don't know the project well enough to confirm it's OK to use it.

garydgregory: commons-lang3, in git master, is a test-only dependency. Personally, I am OK with making it a runtime dependency as well since Commons VFS is a higher level component that already depends on many dependencies.

garydgregory:
@coder-hugo May you please rebase to solve conflicts?

PR #9, opened 2015-06-26
[VFS-497] Ported filters from Commons IO

michael-schnell:
https://issues.apache.org/jira/browse/VFS-497

### DIFF
--- a/core/src/main/java/org/apache/commons/vfs2/filter/AgeFileFilter.java
+++ b/core/src/main/java/org/apache/commons/vfs2/filter/AgeFileFilter.java
@@ -1,0 +1,211 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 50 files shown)


PascalSchumacher:
@ecki @garydgregory Any opinion on this?

garydgregory:
I do not like it. Commons CVS is already a kitchen sink of sorts, so would just centralize all filters in Commons IO and make VFS depend on IO.

garydgregory:
Darn autocomplete! I meant VFS, not CSV.

michael-schnell:
I checked this option before I decided to port the code to VFS. Unfortunately commons IO filter uses the Java type File all over the place. You would need to make an incompatible change to the commons io code or create a new package for it. So, I'd suggest just to merge the (three years old) code.

ecki:
It's a bit heavyweight and I had no need for those filters however I like the fact that the patch does not add a dependency, so it is fine with me. Will merge it.

michael-schnell:
Great. Go ahead!

garydgregory:
Merged with some changes.

PR #12, opened 2016-03-21
avoid deadlock is AbstractFileSystem.resolveFile()

eepstein:
Change to "synchronized" bracketing to avoid a potential deadlock.

### DIFF
--- a/core/src/main/java/org/apache/commons/vfs2/provider/AbstractFileObject.java
+++ b/core/src/main/java/org/apache/commons/vfs2/provider/AbstractFileObject.java
@@ -564,28 +564,24 @@ public abstract class AbstractFileObject<AFS extends AbstractFileSystem> impleme
      */
     private void detach() throws Exception
     {
-        // no need to enter synchronized block if already detached.
-        if (attached)
+        synchronized (fs)
         {
-            synchronized (fs)
+            if (attached)
             {
-                if (attached)
+                try
                 {
-                    try
-                    {
-                        doDetach();
-                    }
-                    finally
-                    {
-                        attached = false;
-                        setFileType(null);
-                        parent = null;
+                    doDetach();
+                }
+                finally
+                {
+                    attached = false;
+                    setFileType(null);
+                    parent = null;
 
-                        // fs.fileDetached(this);
+                    // fs.fileDetached(this);
 
-                        removeChildrenCache();
-                        // children = null;
-                    }
+                    removeChildrenCache();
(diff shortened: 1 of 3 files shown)


eepstein:
Fixed a previously broken test case as well.

core/src/main/java/org/apache/commons/vfs2/provider/AbstractFileSystem.java
```diff
@@ -317,42 +317,46 @@ public FileObject resolveFile(final FileName name) throws FileSystemException
         return resolveFile(name, true);
     }
 
-    private synchronized FileObject resolveFile(final FileName name, final boolean useCache) throws FileSystemException
```
ecki: thats a quite delicate location/change. Can you somehow reproduce the deadlocks or show thread dumps?

core/src/main/java/org/apache/commons/vfs2/provider/AbstractFileObject.java
```diff
@@ -564,24 +564,28 @@ private boolean deleteSelf() throws FileSystemException
      */
     private void detach() throws Exception
     {
-        synchronized (fs)
+        // no need to enter synchronized block if already detached.
```
ecki: this is needed to avoid that the state can change while in the conditional block. It might be possible to use a DCL, but not sure if that helps much. VFS threading/concurrency is not really clear designed in some corners.

eepstein: makes sense.  it was a minor idea, i've reverted that.

core/src/main/java/org/apache/commons/vfs2/provider/AbstractFileSystem.java
eepstein: note that only the last conditional and contained method call is outside the synchronized block.  Alternatively the conditional could be done inside, setting a local variable, and just the refresh() call would be outside the block (though I don't imagine the cache strategy is likely to change mid-flight).  The point being the code is still thread-safe the former still within the synchronized block and the latter is synchronized on the underlying filesystem.

I don't have a reproducible test across multiple threads accessing different files on the same FS that reliably shows the issue.

eepstein: I've heard about http://vmlens.com/ , perhaps that will allow creating a reproducible test case.

PR #13, opened 2016-05-09
VFS-180: minor clean ups on the latest patch in VFS-180 and ...

woonsan:
putting new classes into the same webdav package withouth having to create a new package for webdavs, following Bernd's comment in VFS-180.


PascalSchumacher:
@ecki what is your opinion on this?

PR #14, opened 2016-05-31
reduce heap by interning / lazy init

fommil:
This saves about 75% of the heap in ENSIME after quite a bit of memory profiling. It may increase the PermGen requirements of users on Java 6 but will reduce their overall memory consumption.

We have the ability to run with a monkey patched version of VFS, so we are not in any need of you to merge this, but it may benefit other users.

pinging @rgoers @sebbASF explicitly incase you don't check for github PRs.

Provided under the terms of the Apache 2.0 License.


jvican:
+1

fommil:
No interest, communication only through snail mail. Closing.

rgoers:
I'm sorry. Somehow I missed the email. I haven't actually been doing much on VFS recently, but others have. This really should have been brought up on the commons dev list instead of emailing individuals.

PR #15, opened 2016-06-07
VFS-570 Write Ability for HDFS Provider

garpinc:
VFS-570 Write Ability for HDFS Provider. Push request to include Aron's patch


garpinc:
Hi.. Is there an issue with this pull request or can it be integrated?

PR #17, opened 2016-10-09
[VFS-624] fixed LocalFileRandomAccessContent error in read() for 0xFF

davidgoitia:
Fixed LocalFileRandomAccessContent implementation of InputStream read() method returning wrong negative values as described in: [https://issues.apache.org/jira/browse/VFS-624](https://issues.apache.org/jira/browse/VFS-624)


garydgregory:
Hi,

Thank you for your report. Are you able to provide a unit test patch?

Gary

DasmonBilibili:
The file stream is read according to the byte stream (8bit, 1type), there is no 0xFF

PR #18, opened 2017-03-19
updgrade to jcifs 1.3.17

ilang:
An upgrade from a very old jcifs to almost the new one 
(The newest one is 1.3.18, but is not in the maven repository).


PascalSchumacher:
Thanks! :+1:

PR #19, opened 2017-03-22
Added possibility to set config value which has Object type using Del…

YaraslauBarysenka:
…egatingFileSystemOptionsBuilder class. Reason: in order to initialize FileSystemOptions we should pass AmazonS3Client object for S3 file system.


YaraslauBarysenka:
The Travis CI build failed, but I have committed very simple changes: just added two new set methods. So my changes can not be case of fail build.

PR #23, opened 2017-09-25
Fix sandbox module name

boris-petrov:
Otherwise including the sandbox module in the build fails with an error.


ottobackwards:
This seems correct to me. @garydgregory ?

garydgregory:
Yes, it does. The module name should match both the artifact ID and the folder name.

PR #27, opened 2017-12-17
SFTP: Perform less strict permission checks on servers where the exec channel is disabled

Nimlhug:
Prior to this patch, all permissions checks would fail on servers
without an exec channel: SftpFileObject relied on the knowledge of the
current user's UID and GID. Neither of these can be obtained without an
exec channel. The SFTP specifications do not appear to mention any
alternative method of ascertaining the UID/GID.

As a workaround, when the UID can't be detected, SftpFileObject will use
a DummyPosixPermissions instance, which always "grants" RWX.

There are two sides to this coin:

1. This will allow normal operation against SFTP servers without an exec
channel.
2. This may cause JSchExceptions to be thrown when the actual
permissions are inconsistent with the dummy permissions.

This seems to be preferable to not being able to use exec-less SFTP
servers. Regressions should be limited to these kinds of servers.

Calling getUId when first making a connection might impose an overhead,
but this is presumably a trivial one.

This will probably fix VFS-590 and VFS-617.

This solution could be made a bit more elegant, perhaps by making this check optional, or by using an alternative way to detect whether the exec channel is present.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/DummyPosixPermissions.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/DummyPosixPermissions.java
@@ -1,0 +1,13 @@
+package org.apache.commons.vfs2.provider.sftp;
+
+import org.apache.commons.vfs2.util.PosixPermissions;
+
+/**
+ * Pretends that the current user is always the owner and in the same group.
+ */
+public class DummyPosixPermissions extends PosixPermissions {
+
+    public DummyPosixPermissions(final int permissions) {
+        super(permissions, true, true);
+    }
+}
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
@@ -64,12 +64,18 @@ public class SftpFileSystem extends AbstractFileSystem {
      */
     private int[] groupsIds;
 
+    /**
+     * Some SFTP-only servers disable the exec channel. When exec is disabled, things like getUId() will always fail.
+     */
+    private boolean execDisabled; 
+
     protected SftpFileSystem(final GenericFileName rootName, final Session session,
             final FileSystemOptions fileSystemOptions) {
         super(rootName, null, fileSystemOptions);
         this.session = session;
         final SftpFileSystemConfigBuilder builder = SftpFileSystemConfigBuilder.getInstance();
         this.connectTimeoutMillis = builder.getConnectTimeoutMillis(fileSystemOptions);
+        detectExecDisabled();
     }
 
     @Override
@@ -258,6 +264,13 @@ public class SftpFileSystem extends AbstractFileSystem {
     }
 
     /**
+     * @see SftpFileSystem#execDisabled
+     */
+    public boolean isExecDisabled() {
+        return execDisabled;
+    }
+
+    /**
      * Execute a command and returns the (standard) output through a StringBuilder.
      *
      * @param command The command
@@ -295,5 +308,18 @@ public class SftpFileSystem extends AbstractFileSystem {
         }
         channel.disconnect();
         return channel.getExitStatus();
+    }
+
+    /**
+     * Some SFTP-only servers disable the exec channel.
+     * 
+     * Attempt to detect this by calling getUid.
+     */
+    private void detectExecDisabled() {
+        try {
+            getUId();
+        } catch(JSchException | IOException e) {
+            execDisabled = true;
+        }
     }
 }
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileObject.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileObject.java
@@ -249,6 +249,13 @@ public class SftpFileObject extends AbstractFileObject<SftpFileSystem> {
         statSelf();
         boolean isInGroup = false;
         if (checkIds) {
+            if(getAbstractFileSystem().isExecDisabled()) {
+                // Exec is disabled, so we won't be able to ascertain the current user's UID and GID.
+                // Return "always-true" permissions as a workaround, knowing that the SFTP server won't 
+                // let us perform unauthorized actions anyway.
+                return new DummyPosixPermissions(attrs.getPermissions());
+            }
+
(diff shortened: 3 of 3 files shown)


commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/DummyPosixPermissions.java
```diff
@@ -0,0 +1,13 @@
+package org.apache.commons.vfs2.provider.sftp;
+
+import org.apache.commons.vfs2.util.PosixPermissions;
+
+/**
+ * Pretends that the current user is always the owner and in the same group.
+ */
+public class DummyPosixPermissions extends PosixPermissions {
```
garydgregory: Thank you for your patch. Can you come up with a better class name here? Maybe one that reflects the fact that it "Pretends that the current user is always the owner and in the same group." "Dummy" does not reflect that IMO.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
```diff
@@ -296,4 +309,17 @@ private int executeCommand(final String command, final StringBuilder output) thr
...
+    /**
+     * Some SFTP-only servers disable the exec channel.
+     * 
+     * Attempt to detect this by calling getUid.
+     */
+    private void detectExecDisabled() {
+        try {
+            getUId();
+        } catch(JSchException | IOException e) {
+            execDisabled = true;
```
garydgregory: Shouldn't the exception be stored here so that it can be queries by applications or tests? It would be a pain to have to step into the debugger to figure out what the real failure is here.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
```diff
@@ -296,4 +309,17 @@ private int executeCommand(final String command, final StringBuilder output) thr
...
     }
+
+    /**
+     * Some SFTP-only servers disable the exec channel.
+     * 
+     * Attempt to detect this by calling getUid.
+     */
+    private void detectExecDisabled() {
+        try {
+            getUId();
```
garydgregory: Should this value be cached?

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
Nimlhug: Might be useful. I'm not familiar with the way VFS does error handling/logging. However, an exception here will either indicate the absence of the exec channel (which isn't really a problem and which is what I'm trying to detect), or something serious. In the case of "something serious", the exception (or a similar one) will surface when the SftpFileSystem instance is used.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
Nimlhug: The UID? It probably could be, I can't imagine it changing at runtime. But it wasn't cached and I figured I'd keep my patch as small as possible.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/DummyPosixPermissions.java
Nimlhug: LoggedInUserRWXPermissions? UserGroupImpersonatingPermissions? Open to suggestions here, I struggled and came up with dummy ;-)

henjovr: Simply _IsOwnerInGroupPosixPermissions_ ?

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
henjovr: Log on debug such that we can always turn on debug logging for VFS to get exceptions.
( log.debug(....); )

henjovr:
This PR is very useful for handling SFTP connections against crappy SSHD/SFTP servers without exec channels. There are unfortunately plenty of those :(

I'd really love to see this included in 2.2.1, since I'm running it already succesfully with a custom built version of VFS 2.2.1-SNAPSHOT.

Nimlhug:
I'll try to process most of this feedback tomorrow and update the PR. Apologies for the lacking reactivity, been running this in production without issues so managed to forget all about it :-)

Nimlhug:
I've processed the feedback. Please let me know if there's anything else you need for me in order to get this accepted.

dmcg:
Hey, any progress on this please? It's blocking deployment for us

garydgregory:
The patch is missing unit tests.

Gary

ajrnz:
As a possible workaround if you don't want to maintain a fork or wait for this PR, call the `doRename()` method of `SftpFileObject` using reflection. (This is scala rather than java but the changes are trivial)

```scala
val doRenameMethod = classOf[SftpFileObject].getDeclaredMethod("doRename", classOf[FileObject])
doRenameMethod.setAccessible(true)
doRenameMethod.invoke(file, newFile)
```
Obviously you need to check the object type and make sure you understand what the code normally does

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
```diff
@@ -296,4 +314,18 @@ private int executeCommand(final String command, final StringBuilder output) thr
...
+        } catch(JSchException | IOException e) {
+            execDisabled = true;
```
garydgregory: How about a unit test?

jclx:
I see this was closed. I couldn't find anything else about this issue but our sftp servers disallows exec channel calls likes 'id -G' and returns this string "This service allows sftp connections only." and then fails with an exception and thus can't do any calls that use isWritable,isReadable, isExecutable. vfs sftp is unusable. It there anything in the works or an issue for this so vfs can handle this case at a future date time.
Current workaround is to create our own SFTP Provider and subclass and disable exec commands.

markt-asf:
It was closed automatically when the obsolete trunk branch was deleted. I'd re-open the PR but the source repo has been deleted so I can't. If this is still a concern, please open a new PR against master.

PR #28, opened 2018-02-01
fixed for AbstractFileObject#toString leak password!

qxo:
change  fileName.getURI() to fileName.getFriendlyURI() for improve security( mask password )


garydgregory:
Any thoughts on the likelihood of this breaking existing apps?

PR #29, opened 2018-02-16
VFS-398 FtpFileObject.getChildren() fails when a folder contains a file with a colon in the name

ottobackwards:
> This issue is not provider dependent

The issue here is that it is possible for file names to have `:` in them for some providers ( such as FTP).  This causes an issue
because the code wants evaluate a string which may be a name or an url with a scheme to extract, and does so by
searching for `:` and assuming everything to the left is the scheme.

### From Jira:
The code wants to support a string the may be a url or may not be at the same time.
The issue is, that no matter what we do we can never know if the ':' is the scheme marker or part of the file name.
each scheme by definition supports a different scheme part ( the part after the : ) and some don't even have one.
Add to this, that the scheme is used in the system to map to a provider, and those can be registered with whatever scheme a developer wants.

My proposal is to limit the recognized scheme's returned by extractScheme to those registered in the system.  I believe this is the most correct way to do this.
- it is only way to be sure that it is a scheme and not just a ':'
- the system can only support registered schemes anyway

-----------

So this PR introduces a new method to the URLParser which will extract the scheme using comparison to the schemes registered in the system currently.

### Questions for review
- Is the scope of this fix sufficient, or does the extractSchema() call need to be completely removed and replaced and all access changed?  ( I think this will need to be done )
- If a public method is removed and replaced, how does this project handle that?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ target/
 .classpath
 .settings/
 .svn/
+.idea/
+*.iml
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/impl/DefaultFileSystemManager.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/impl/DefaultFileSystemManager.java
@@ -175,7 +175,7 @@ public class DefaultFileSystemManager implements FileSystemManager {
      * @throws FileSystemException if an error occurs adding the provider.
      */
     public void addProvider(final String urlScheme, final FileProvider provider) throws FileSystemException {
-        addProvider(new String[] { urlScheme }, provider);
+        addProvider(new String[] {urlScheme}, provider);
     }
 
     /**
@@ -346,7 +346,7 @@ public class DefaultFileSystemManager implements FileSystemManager {
         }
 
         try {
-            fileObjectDecoratorConst = fileObjectDecorator.getConstructor(new Class[] { FileObject.class });
+            fileObjectDecoratorConst = fileObjectDecorator.getConstructor(new Class[] {FileObject.class});
         } catch (final NoSuchMethodException e) {
             throw new FileSystemException("vfs.impl/invalid-decorator.error", fileObjectDecorator.getName(), e);
         }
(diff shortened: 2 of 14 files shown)


ottobackwards:
I have left the old version of the method for comparison, I think it should be removed before the end of review

ottobackwards:
Another reason I didn't remove the old function is that I don't know how this project handles possible breaking changes?

gbulfon:
Hi, did any patch to the colon bug ever reached a vfs2 release?

ottobackwards:
No, 

- no review as of yet
- there are no 'releases' of vfs planned that I can see.

Prs are not getting reviewed

gbulfon:
what if we make a fork into our Sonicle repositories and put your changes?

ottobackwards:
That is always your option to do.  You'll have to consider the risks and the work you'll need to do to re-integrate when and if it lands.

Also, I would think you would want to kick the tires and test it out of course.

You can, if you find issues, comment on this pr..

pkcinna01:
Can a quick summary be given on how the system gets the initial supported schemes?  I'm wondering if having the new method with list of supported schemes is flexible enough and if one more layer is needed.  For example, a SupportedSchemeListMatcher would be default but you could make your own by extending some interface.  I know class name above is not good but just throwing this out there to help clarify idea.  On the other hand the fix already proposed should address the issue.

ottobackwards:
The only schemes that can ever be supported are the schemes for the existing providers loaded in the system.  The DefaultFileSystemManager now sends those in with the request.

There are no viable 'alternative' schemes.

The question I have, is if changing the default provider is enough.

ottobackwards:
I'll do the java doc.  Thanks @garydgregory

ottobackwards:
@garydgregory I added the javadoc.  I would almost want to @depricate the other methods... but I'm not sure.   This is a static api not an interface.....

garydgregory:
You can deprecate anything, it does not have to be a VFS interface. We cannot break anything that is ```public```; unless it's in an internal package or class, but we do have these in VFS ATM.

ottobackwards:
@garydgregory hopefully the latest is suitable. It is consistent with other such messages

ottobackwards:
@garydgregory anything else I need to do here?

ottobackwards:
I am not sure why travis is failing. 
`mvn clean install site:site` is working locally.  I'm going to kick it.

jgribonvald:
Could you provide a snapshot with these change to be able to test ?

ottobackwards:
@jgribonvald you mean a jar?  You build my branch from my repo?

ottobackwards:
I have no way of posting a jar, you'll have to build it

PR #32, opened 2018-03-19
Allow to set key exchange algorithm explicitly

KedemBarak:


garydgregory:
Hi,

I do not see the code changes in git master, so it will not be in 2.3 which
is going through validation ATM.

You should create a Jira ticket so make this request more visibility and
link to the PR. I did only looked at Jira when I went through tickets for
2.3. We use Jira for issue tracking, not GitHub, I am sorry if that is
confusing.

Gary

PR #36, opened 2018-04-24
fix for VFS-662: SftpFileSystem has Thread-safe issue about idleChannel

qxo:
ref: https://issues.apache.org/jira/browse/VFS-662

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
@@ -50,7 +50,7 @@ public class SftpFileSystem extends AbstractFileSystem {
 
     // private final JSch jSch;
 
-    private ChannelSftp idleChannel;
+    private volatile ChannelSftp idleChannel;
 
     private final int connectTimeoutMillis;
 
@@ -96,11 +96,16 @@ public class SftpFileSystem extends AbstractFileSystem {
         ensureSession();
         try {
             // Use the pooled channel, or create a new one
-            final ChannelSftp channel;
+            ChannelSftp channel = null;
             if (idleChannel != null) {
-                channel = idleChannel;
-                idleChannel = null;
-            } else {
+                synchronized (this) {
+                    if (idleChannel != null)  {
+                        channel = idleChannel;
+                        idleChannel = null;
+                    }
+                }
+            }
+            if ( channel == null ) {
                 channel = (ChannelSftp) session.openChannel("sftp");
                 channel.connect(connectTimeoutMillis);
                 final Boolean userDirIsRoot = SftpFileSystemConfigBuilder.getInstance()
@@ -139,7 +144,7 @@ public class SftpFileSystem extends AbstractFileSystem {
      */
     private void ensureSession() throws FileSystemException {
         if (this.session == null || !this.session.isConnected()) {
-            doCloseCommunicationLink();
+            closeCommunicationLink();
 
             // channel closed. e.g. by freeUnusedResources, but now we need it again
             Session session;
@@ -172,10 +177,17 @@ public class SftpFileSystem extends AbstractFileSystem {
      */
     protected void putChannel(final ChannelSftp channel) {
         if (idleChannel == null) {
-            // put back the channel only if it is still connected
-            if (channel.isConnected() && !channel.isClosed()) {
-                idleChannel = channel;
-            }
+           synchronized (this) {
+               if (  idleChannel == null ) {
+                    // put back the channel only if it is still connected
+                    if (channel.isConnected() && !channel.isClosed())
+                    {
+                        idleChannel = channel;
+                    }
+               }else{
+                    channel.disconnect();
+               }
+           }
         } else {
             channel.disconnect();
         }


garydgregory:
Thank you for your patch. Can you add comments to your code to explain why this is a problem and how it is fixed? Also, can you explain if it is really better to synchronize these blocks instead of the whole method? IOW, are there other race conditions that would be addressed by synchronizing the method?

qxo:
It's suit for my RT env. 
 synchronize  code only if it need, so I not sync method:)

garydgregory:
@qxo ,

I looked at this today and was reminded that DCL is broken: https://www.javaworld.com/article/2074979/double-checked-locking--clever--but-broken.html

I will consider this some more but there are two alternatives: synchronize `getChannel()` and `putChannel()` or, more simply, use eager initialization of the `idleChannel` ivar.

_To be continued..._

boris-petrov:
@garydgregory - DCL has not been broken for [a long time now](https://stackoverflow.com/a/3580658/4478636) as long as you use a `volatile` instance variable. :)

garydgregory:
Hi All:

Combining DCL and volatile still does not get it right in the current state of this patch since the `Session` is not volatile. While I appreciate the pragmatic approach this patch takes, the fact that it works _for now_ for @qxo is not the complete solution IMO.

Would it not be simpler and therefore less bug prone and easier to maintain to synchronize the get and put methods? Granted this is more expensive. Providing finer grain locking as this patch attempts needs to account for all cases, which would be nice to have.

Thoughts?

@qxo : May you rebase this patch on master please?

boris-petrov:
@garydgregory - I'm not familiar with the specific code here but the diff that I see here is totally safe - `idleChannel` is `volatile` and it's the only variable on which DCL is performed. This is safe since Java 1.5. Why do you think `session` also has to be?

ottobackwards:
I'm just going to raise my hand and say from what I've seen,  I didn't think VFS was thread safe.  Even though some bugs like this have been fixed from time to time, in the main I don't think it is is it?

boris-petrov:
@ottobackwards - well, I've been using VFS for a while now in a multi-threaded environment and it mostly works fine. There are some issues here and there but I think they will be resolved one by one and all will be good in the end.

garydgregory:
I'll second @boris-petrov's comment. We've been patching problems here and there when they pop up. I'm sure there are plenty of MT issues here and there. YMMV with each provider. The core probably has it's own issues too.

boris-petrov:
@garydgregory, @ottobackwards - in that regards, please take a look at [this PR](https://github.com/apache/commons-vfs/pull/54) which is by @abashev. In his S3 provider he uses a fork of VFS in which he claims he's fixed a number of concurrency issues. He wants to merge them in the upstream repo (this one) so please take a look at the PR I linked (and accept it/comment on it) so he can continue on with the real issues he thinks he's resolved.

garydgregory:
Hi All:

Back to https://github.com/apache/commons-vfs/pull/36#issuecomment-487393002:

The `session` ivar is not final and gets written in the private method `ensureSession()`, so it is possible for it to get initialized multiple times in a race-condition. The method `getChannel()` can end up writing to both `session` and `idleChanne`, so both values need to be published to other threads. The simplest way to do this is to synchronize the `getChannel()` method.

Thoughts?

garydgregory:
> @garydgregory, @ottobackwards - in that regards, please take a look at [this PR](https://github.com/apache/commons-vfs/pull/54) which is by @abashev. In his S3 provider he uses a fork of VFS in which he claims he's fixed a number of concurrency issues. He wants to merge them in the upstream repo (this one) so please take a look at the PR I linked (and accept it/comment on it) so he can continue on with the real issues he thinks he's resolved.

I do not see anything in that PR about concurrency. That PR just fiddles with POMs. What am I missing?

boris-petrov:
> The `session` ivar is not final and gets written in the private method `ensureSession()`, so it is possible for it to get initialized multiple times in a race-condition. The method `getChannel()` can end up writing to both `session` and `idleChanne`, so both values need to be published to other threads. The simplest way to do this is to synchronize the `getChannel()` method.
> 
> Thoughts?

Well, in that case there is still a race-condition because of the `executeCommand` method which also calls `ensureSession` so I guess it also should be `synchronized`...?

> I do not see anything in that PR about concurrency. That PR just fiddles with POMs. What am I missing?

This PR has nothing to do with concurrency but as I said @abashev's fork of VFS has some fixes in it which he wants to merge - and this PR is just the first one of a few to come as far as I understand.

garydgregory:
Well, let just deal with `SftpFileSystem` here... we can await for other PRs to come in...

Gary

garydgregory:
So, in the DCL+volatile dept, I did this: 

https://github.com/apache/commons-vfs/blob/master/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java#L220-L242

In this case, a single ivar is in play in one place.

abashev:
> This PR has nothing to do with concurrency but as I said @abashev's fork of VFS has some fixes in it which he wants to merge - and this PR is just the first one of a few to come as far as I understand.

@garydgregory Yes, it will be my first submission into commons-vfs project and I want to understand how it works and the whole process. It looks like pom file was changed so I did rebase on latest master for this one https://github.com/apache/commons-vfs/pull/54 and add one more to fix Java 11 builds https://github.com/apache/commons-vfs/pull/61

garydgregory:
BTW, welcome aboard the Apache Commons project :-)

PR #39, opened 2018-08-28
[VFS-673] Support com.jcraft.jsch.ConfigRepository (~/.ssh/config) with SftpFileSystemConfigBuilder and flag to load OpenSSHConfig

lihonosov:
Allow to set com.jcraft.jsch.ConfigRepository explicitly and flag to load OpenSSHConfig if the ConfigRepository was not provided.

[JSch-Examples-OpenSSHConfig](http://www.jcraft.com/jsch/examples/OpenSSHConfig.java.html)


lihonosov:
The change is in svn
https://github.com/apache/commons-vfs/commit/ae88cc36f788f41365b8e326fb0774fcfcad9f7a

PR #41, opened 2018-11-06
Fix throwing exception on normal FTP stream close

boris-petrov:
Fixes https://issues.apache.org/jira/projects/VFS/issues/VFS-674.

For some reason, some FTP servers return `426` when you abort a download request before it has completed. You can see this happening with a FileZilla server:

`curl -v --range 0-499 ftp://ffmpeg:ffmpeg2018@ftp.pbteu.com/ExportHD.mpg --output -`

With the output being:

```
...
> RETR ExportHD.mpg
< 150 Opening data channel for file download from server of "/ExportHD.mpg"
* Maxdownload = 500
* Getting file with size: 500
* Remembering we are in dir ""
> ABOR
< 426 Connection closed; transfer aborted.
...
```

This PR handles cases like this.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ftp/FtpFileObject.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ftp/FtpFileObject.java
@@ -29,6 +29,7 @@ import java.util.TreeMap;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.commons.net.ftp.FTPFile;
+import org.apache.commons.net.ftp.FTPReply;
 import org.apache.commons.vfs2.FileName;
 import org.apache.commons.vfs2.FileNotFolderException;
 import org.apache.commons.vfs2.FileObject;
@@ -582,7 +583,8 @@ public class FtpFileObject extends AbstractFileObject<FtpFileSystem> {
         protected void onClose() throws IOException {
             final boolean ok;
             try {
-                ok = client.completePendingCommand();
+                // See VFS-674 and the accompanying PR as to why this check for "transfer aborted" is needed
+                ok = client.completePendingCommand() || client.getReplyCode() == FTPReply.TRANSFER_ABORTED;
             } finally {
                 getAbstractFileSystem().putClient(client);
             }
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ftp/FTPClientWrapper.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ftp/FTPClientWrapper.java
@@ -264,6 +264,11 @@ public class FTPClientWrapper implements FtpClient {
     }
 
     @Override
+    public int getReplyCode() throws IOException {
+        return getFtpClient().getReplyCode();
+    }
+
+    @Override
     public String getReplyString() throws IOException {
         return getFtpClient().getReplyString();
     }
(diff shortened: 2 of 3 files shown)


boris-petrov:
> Hm, this does mean you asume ok when the server tells you you had a incomplete download?

No, this is only meant to handle the client telling the server to interrupt the connection.

> Maybe an "consume rest" method would be better for those who cant ignore the exception on close?

Well, most third-party code doesn't handle such a case. Why would anyone handle an IOException when you're closing the stream?

> It is also a security problem if somebody can truncate SSL sessions and stuff...

Not sure what you mean here?

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ftp/FtpFileObject.java
```diff
@@ -582,7 +582,7 @@ void abort() throws IOException {
         protected void onClose() throws IOException {
             final boolean ok;
             try {
-                ok = client.completePendingCommand();
+                ok = client.completePendingCommand() || client.getReplyCode() == 426;
```
garydgregory: Whether or not this fix is right, you should use the constant ```FTPReply.TRANSFER_ABORTED```, not a magic number.

garydgregory: Also, are there any other return codes that should be considered? Should a list of such RCs be configurable? That last part seems over the top.

boris-petrov: Thanks for the comments! I've updated the PR code to use the constant instead.

As for the second part - I'm still not sure whether other return codes are possible in "normal" circumstances. I'll be debugging in the next few days. But I think this is enough for now, no need to over-design this from the start. If there are other cases like this one, we could always make more changes later.

ecki:
Ok I guess VFS is a reasonable high abstraction layer that users would not care about getting informed about such nuances from servers, so I guess we can swallow the abort message. (Maybe we could however detect if the input stream thinks we have read to the end and if we did, then not ignore it?)

My other content about TLS was that to avoid[ truncation attacks](https://en.wikipedia.org/wiki/Transport_Layer_Security#Truncation_attack) you typically have to expect a "close notify" and if that is not coming you would normally return an error on close (or on a synthetic last read).

The problem is that a SSLSocket EOFException which would notify you about the unclean shutdown is swallowed because the server also alerts you about the unclean shutdown. Thats a bit sad. So when I know I stop reading and clsoe a socket I do expect possible exceptions from the close even on read. But I know I might be special about that.

boris-petrov:
@ecki 

> Ok I guess VFS is a reasonable high abstraction layer that users would not care about getting informed about such nuances from servers, so I guess we can swallow the abort message. 

Yes, that is also my thinking.

> (Maybe we could however detect if the input stream thinks we have read to the end and if we did, then not ignore it?)

Well, it's not possible to know if the stream has been fully read so not sure how we can do this?

> Is the onClose also used for OutputStreams? Maybe it should not ignore it in that case....

It is not used, no. There is another class for that and I haven't touched it there.

> My other content about TLS

I'm not that familiar with the issue you describe. But TLS is a layer on top of the FTP protocol and these return codes are specific to it, I'm not sure this is a relevant problem here... but I'm by no means an expert.

boris-petrov:
Thank you for merging this patch! I don't see, however, the change in `FTPClientWrapper` so `getReplyCode` in `FtpClient` will always return `FTPReply.COMMAND_OK` or am I missing something?

boris-petrov:
Oh, OK, ignore what I said, I saw that you've done this in [another commit](https://github.com/apache/commons-vfs/commit/5acc25cdd504a2c65ce993a22e999eb2b19995b8). Thank you very much for the support! I'm looking forward to the 2.3 release. :)

PR #43, opened 2018-11-16
setManager should be synchronized

RobertDeRose:
Since `getManager` is synchronized to protect the `instance` resource,
so should the `setManager` since it operates on the same resource.


RobertDeRose:
ping @garydgregory

garydgregory:
Hm, would that really do anything in this case?

PR #44, opened 2018-11-26
More efficient comparison in FileExtensionSelector

RobertDeRose:
* Removed duplication from Constructors
* Removed unneeded check for `null` `HashSet` as it is constructed during object creation by the JVM at runtime.
* Optimized the comparison over the Constructor by converting all extensions to lowercase and using the `contains` method of a `Set`.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/FileExtensionSelector.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/FileExtensionSelector.java
@@ -47,7 +47,11 @@ public class FileExtensionSelector implements FileSelector {
      */
     public FileExtensionSelector(final Collection<String> extensions) {
         if (extensions != null) {
-            this.extensions.addAll(extensions);
+            for (String ext : extensions) {
+                if (ext != null) {
+                    this.extensions.add(ext.toLowerCase());
+                }
+            }
         }
     }
 
@@ -57,9 +61,7 @@ public class FileExtensionSelector implements FileSelector {
      * @param extensions The extensions to be included by this selector.
      */
     public FileExtensionSelector(final String... extensions) {
-        if (extensions != null) {
-            this.extensions.addAll(Arrays.asList(extensions));
-        }
+        this(Arrays.asList(extensions));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


RobertDeRose:
@garydgregory Any thoughts on this?

commons-vfs2/src/main/java/org/apache/commons/vfs2/FileExtensionSelector.java
```diff
@@ -47,7 +47,11 @@
      */
     public FileExtensionSelector(final Collection<String> extensions) {
         if (extensions != null) {
-            this.extensions.addAll(extensions);
+            for (String ext : extensions) {
+                if (ext != null) {
```
garydgregory: +1 to the null check.

commons-vfs2/src/main/java/org/apache/commons/vfs2/FileExtensionSelector.java
```diff
@@ -57,9 +61,7 @@ public FileExtensionSelector(final Collection<String> extensions) {
      * @param extensions The extensions to be included by this selector.
      */
     public FileExtensionSelector(final String... extensions) {
-        if (extensions != null) {
-            this.extensions.addAll(Arrays.asList(extensions));
-        }
+        this(Arrays.asList(extensions));
```
garydgregory: +1 to reusing the ctor.

commons-vfs2/src/main/java/org/apache/commons/vfs2/FileExtensionSelector.java
```diff
@@ -70,15 +72,7 @@ public FileExtensionSelector(final String... extensions) {
...
-        if (this.extensions == null) {
-            return false;
-        }
-        for (final String extension : this.extensions) {
-            if (fileInfo.getFile().getName().getExtension().equalsIgnoreCase(extension)) {
-                return true;
-            }
-        }
-        return false;
+        return this.extensions.contains(fileInfo.getFile().getName().getExtension().toLowerCase());
```
garydgregory: I'm not sure that's quite right. Unit tests would help check that case is handled properly. Note that not all file systems are case-insenstive. OTOH, if that is an issue with this code, it was already there! ;-)

PR #47, opened 2018-12-07
VFS-685 Modify FileInfo test class so it is usable in outside projects

ottobackwards:
Outside projects may not only run, but may extend the test classes in the project.  They cannot do so without the FileInfo class being public.


ottobackwards:
@garydgregory can we get this going?

PR #51, opened 2019-02-23
Add support for customizing FTP transfer aborted status codes

boris-petrov:
This is connected with [another PR I did a few months ago](https://github.com/apache/commons-vfs/pull/41). We hit yet **another** FTP that returns some strange status code when prematurely closing a stream - this one returns `550`.

This PR makes these status codes that should be treated as *OK* configurable with a default of `426` and `550` to handle such cases like the ones we've encountered.

P.S. Here is output from a `curl -v --range 0-499 "ftp://user:pass@192.168.40.224/dir/file.mpg" --output -` command:

```
...
> RETR file.mpg
< 125 Data connection already open; Transfer starting.
* Maxdownload = 500
* Getting file with size: 500
* Remembering we are in dir "dir/"
> ABOR
< 550 Data channel was closed by ABOR command from client.
* partial download completed, closing connection
> QUIT
< 226 ABOR command successful.
* Closing connection 0
```

P.S.2 This happens on a "Microsoft FTP Service":

```
...
220 Microsoft FTP Service
...
```

This is from the curl log.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ftp/FtpFileSystemConfigBuilder.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ftp/FtpFileSystemConfigBuilder.java
@@ -17,7 +17,11 @@
 package org.apache.commons.vfs2.provider.ftp;
 
 import java.net.Proxy;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
+import org.apache.commons.net.ftp.FTPReply;
 import org.apache.commons.net.ftp.parser.FTPFileEntryParserFactory;
 import org.apache.commons.vfs2.FileSystem;
 import org.apache.commons.vfs2.FileSystemConfigBuilder;
@@ -46,6 +50,7 @@ public class FtpFileSystemConfigBuilder extends FileSystemConfigBuilder {
     private static final String SHORT_MONTH_NAMES = _PREFIX + ".SHORT_MONTH_NAMES";
     private static final String SO_TIMEOUT = _PREFIX + ".SO_TIMEOUT";
     private static final String USER_DIR_IS_ROOT = _PREFIX + ".USER_DIR_IS_ROOT";
+    private static final String TRANSFER_ABORTED_OK_STATUS_CODES = _PREFIX + ".TRANSFER_ABORTED_OK_STATUS_CODES";
 
     private FtpFileSystemConfigBuilder() {
         super("ftp.");
@@ -73,6 +78,11 @@ public class FtpFileSystemConfigBuilder extends FileSystemConfigBuilder {
     @Override
     protected Class<? extends FileSystem> getConfigClass() {
         return FtpFileSystem.class;
+    }
+
+    public static List<Integer> getSaneTransferAbortedOkStatusCodes() {
+        // See VFS-674, its accompanying PR and https://github.com/apache/commons-vfs/pull/51 as to why 426 and 550 are here
+        return new ArrayList<>(Arrays.asList(FTPReply.TRANSFER_ABORTED, FTPReply.FILE_UNAVAILABLE));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ftp/FtpFileSystemConfigBuilder.java
```diff
@@ -238,6 +242,19 @@ public Boolean getUserDirIsRoot(final FileSystemOptions opts) {
...
+     * @param opts The FileSystem options.
+     * @return The list of status codes (apart from 200) that are considered as OK when prematurely
+     * closing a stream. Defaults to <code>[426, 550]</code>.
+     * @since 2.4
+     */
+    public List<Integer> getTransferAbortedOkStatusCodes(final FileSystemOptions opts) {
+        // See VFS-674, its accompanying PR and https://github.com/apache/commons-vfs/pull/51 as to why 426 and 550 are here
+        @SuppressWarnings("unchecked")
+        List<Integer> statusCodes = (List<Integer>) getParam(opts, TRANSFER_ABORTED_OK_STATUS_CODES);
+        return statusCodes != null ? statusCodes : Arrays.asList(FTPReply.TRANSFER_ABORTED, FTPReply.FILE_UNAVAILABLE);
```
garydgregory: @boris-petrov ,

Thank you so much for your PR Boris. I think you need a unit test that sets expectations for how this configuration option works. For example *I* was expecting this to work but it will not:

```getTransferAbortedOkStatusCodes(opts).add(555)```

I think the above should be fixed which implies a different implementation in the builder. Perhaps the builder is initialized with ```Arrays.asList(FTPReply.TRANSFER_ABORTED, FTPReply.FILE_UNAVAILABLE);``` if it is null. IOW, the new array should be saved.

garydgregory: @boris-petrov,

One more detail.

I would also create a protected method called ```getDefault...``` that returns the ```Arrays.asList(FTPReply.TRANSFER_ABORTED, FTPReply.FILE_UNAVAILABLE)``` to make it clear what the defaults are with the possibility of easy override though subclassing. 

Gary

garydgregory:
@boris-petrov 

And another other thing ;-) What about FTPS?

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ftp/FtpFileSystemConfigBuilder.java
boris-petrov: Thanks for the review! I'm personally against mutable containers in configuration options but if you think that's the way to go, I'll do it, no problem.

I'll do it with the `getDefault...` method as you suggest. I'll also try to add some unit tests.

boris-petrov:
Not sure about FTPS - I guess the same issues apply there, correct? I see that `FtpsFileSystemConfigBuilder` inherits from `FtpFileSystemConfigBuilder` so all should be fine or am I missing something?

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ftp/FtpFileSystemConfigBuilder.java
boris-petrov: Actually, now that I think about it, I'll leave the default value to be `null`. People can specify their own values if they need. Is that fine? And in that case I guess `getDefault...` method is not needed?

boris-petrov:
@garydgregory - I force-pushed. Please check it out. I've left a default value of `null` and added a helper `getSaneTransferAbortedOkStatusCodes` that people can use. Is that fine or you prefer to have these values as default after all?

garydgregory:
@boris-petrov , 

Also, please do not forget to add an entry in `changes.xml`.

Gary

boris-petrov:
@garydgregory - I added an entry in `changes.xml`.

boris-petrov:
@garydgregory - thanks a lot for merging this! Just a note - I see in the commit you renamed "Status Code" to "Reply Code" but you've missed a few places. Just CTRL+F and search for "status" on the [commit page](https://github.com/apache/commons-vfs/commit/044a007664a65ba56539aa098ac1467a4a32b3b3) - there are a couple in `isTransferAbortedOkReplyCode`, `getSaneTransferAbortedOkStatusCodes` and in the JavaDoc comments.

Thanks again!

garydgregory:
@boris-petrov Thank you for reviewing the changes. I just pushed another round of changes.

garydgregory:
@boris-petrov,

My bad for missing those. Thank you again for reviewing.

Gary

boris-petrov:
@garydgregory - thanks! Still one more place left though - in `setTransferAbortedOkReplyCodes` - the method parameter and the docs. :)

garydgregory:
@boris-petrov How about now? ;-)

boris-petrov:
@garydgregory - perfect! :) Thank you very much again! Any chance of a minor release some time soon? :)

garydgregory:
Well, considering we just released 2.3, we might want to accumulate a few more changes before releasing again.

PR #54, opened 2019-03-15
Use parent group ids for all artifacts

abashev:
Hi, I'm the main contributor for this project https://github.com/abashev/vfs-s3 I have a lot of complaints about our 'custom' build for commons-vfs project so I've decided to submit everything that I have to common repo.
There is my first one - using parent group ids for all artifacts. I have to prepare own builds for commons-vfs so for me it is hard to track a lot of places with my group id. With this commit, I will be able to change only parent group id and leave everything as is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ target/
 *.iml
 *.iws
 .DS_Store
+
+.vscode/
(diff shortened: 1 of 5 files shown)


ecki:
I am not a big fan of it as you cannot easily work on the core module but I also won’t oppose it. Could you squash your commits.

abashev:
@ecki done

abashev:
I did rebase for latest master

boris-petrov:
@garydgregory, @ecki - could you take a look/merge this so @abashev can continue with his PRs? :)

PR #55, opened 2019-03-21
Fix inability to start the DefaultFileMonitor after it has been stopped

boris-petrov:


ottobackwards:
Can you add a test for this?

boris-petrov:
Oh, I just force-pushed and you merged it before that... :) I'll create a new PR.

PR #59, opened 2019-04-24
Add ability to specify buffer sizes

boris-petrov:
Fixes https://issues.apache.org/jira/projects/VFS/issues/VFS-705.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/FileContent.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/FileContent.java
@@ -139,6 +139,7 @@ public interface FileContent extends Closeable {
      * Returns an input stream for reading the file's content.
      * <p>
      * There may only be a single input or output stream open for the file at any time.
+     * </p>
      *
      * @return An input stream to read the file's content from. The input stream is buffered, so there is no need to
      *         wrap it in a {@code BufferedInputStream}.
@@ -146,6 +147,23 @@ public interface FileContent extends Closeable {
      *             opening the stream.
      */
     InputStream getInputStream() throws FileSystemException;
+
+    /**
+     * Returns an input stream for reading the file's content.
+     * <p>
+     * There may only be a single input or output stream open for the file at any time.
+     * </p>
+     *
+     * @param bufferSize The buffer size to use.
+     * @return An input stream to read the file's content from. The input stream is buffered, so there is no need to
+     *         wrap it in a {@code BufferedInputStream}.
+     * @throws FileSystemException If the file does not exist, or is being read, or is being written, or on error
+     *             opening the stream.
+     * @since 2.4
+     */
+    default InputStream getInputStream(int bufferSize) throws FileSystemException {
+        return getInputStream();
+    }
 
     /**
      * Returns an output stream for writing the file's content.
@@ -199,6 +217,7 @@ public interface FileContent extends Closeable {
      * <p>
      * This method is a hint to the implementation that it can release resources. This object can continue to be used
      * after calling this method.
+     * </p>
      *
      * @throws FileSystemException if an error occurs closing the file.
      */
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/FileContent.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/FileContent.java
@@ -170,8 +170,10 @@ public interface FileContent extends Closeable {
      * <p>
      * If the file does not exist, this method creates it, and the parent folder, if necessary. If the file does exist,
      * it is replaced with whatever is written to the output stream.
+     * </p>
      * <p>
      * There may only be a single input or output stream open for the file at any time.
+     * </p>
      *
      * @return An output stream to write the file's content to. The stream is buffered, so there is no need to wrap it
      *         in a {@code BufferedOutputStream}.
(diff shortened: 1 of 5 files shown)


commons-vfs2/src/main/java/org/apache/commons/vfs2/FileContent.java
```diff
@@ -147,6 +147,19 @@
...
+     * There may only be a single input or output stream open for the file at any time.
+     *
+     * @param bufferSize The buffer size to use.
+     * @return An input stream to read the file's content from. The input stream is buffered, so there is no need to
+     *         wrap it in a {@code BufferedInputStream}.
+     * @throws FileSystemException If the file does not exist, or is being read, or is being written, or on error
+     *             opening the stream.
+     */
+    InputStream getInputStream(int bufferSize) throws FileSystemException;
+
```
garydgregory: I think you need to make this a default method to avoid breaking compatibility.

commons-vfs2/src/main/java/org/apache/commons/vfs2/FileContent.java
```diff
@@ -194,6 +207,23 @@
...
+     *
+     * @param bAppend true if you would like to append to the file. This may not be supported by all implementations.
+     * @param bufferSize The buffer size to use.
+     * @return An output stream to write the file's content to. The stream is buffered, so there is no need to wrap it
+     *         in a {@code BufferedOutputStream}.
+     * @throws FileSystemException If the file is read-only, or is being read, or is being written, or bAppend is true
+     *             and the implementation does not support it, or on error opening the stream.
+     */
+    OutputStream getOutputStream(boolean bAppend, int bufferSize) throws FileSystemException;
+
```
garydgregory: I think you need to make this a default method to avoid breaking compatibility.

commons-vfs2/src/main/java/org/apache/commons/vfs2/util/MonitorInputStream.java
```diff
@@ -34,6 +34,10 @@ public MonitorInputStream(final InputStream in) {
         super(in);
     }
 
+    public MonitorInputStream(final InputStream in, final int size) {
```
garydgregory: Please add a Javadoc comment for this public method.

commons-vfs2/src/main/java/org/apache/commons/vfs2/util/MonitorOutputStream.java
```diff
@@ -33,6 +33,10 @@ public MonitorOutputStream(final OutputStream out) {
         super(out);
     }
 
+    public MonitorOutputStream(final OutputStream out, final int size) {
```
garydgregory: Please add a Javadoc comment for this public method.

garydgregory:
See comments in individual farms.
The new public methods are missing unit tests.

commons-vfs2/src/main/java/org/apache/commons/vfs2/util/MonitorInputStream.java
boris-petrov: This is and the other one are done.

commons-vfs2/src/main/java/org/apache/commons/vfs2/FileContent.java
boris-petrov: I'm not sure how to approach this... what would the default method do? Throw an `UnsupportedOperationException`? Or call the original method by just ignoring the argument?

boris-petrov:
I'm not sure what kind of tests to write. This is pretty much not adding any new functionality, just exposing an API that Java already has. Any ideas?

garydgregory:
> I'm not sure what kind of tests to write. This is pretty much not adding any new functionality, just exposing an API that Java already has. Any ideas?

Suggestions for tests:
- Validate input:
-- What happens when I pass in a negative number?
-- What happens when I pass in 0?
- Do I really get a buffer sized N? IOW, is the parameter used at all?

commons-vfs2/src/main/java/org/apache/commons/vfs2/FileContent.java
garydgregory: > Or call the original method by just ignoring the argument?

IMO, yes.

boris-petrov: Done.

boris-petrov:
OK, I pushed a couple of tests. They depend on the fact that the BufferedInput/OutputStream constructors check the `size` argument and throw an exception if it is <= 0. Unfortunately I can't check the buffer size as it is a protected variable in the class (unless we expose it in the `MonitorInput/OutputStream` classes which would be ugly).

commons-vfs2/src/main/java/org/apache/commons/vfs2/FileContent.java
```diff
@@ -147,6 +147,22 @@
      */
     InputStream getInputStream() throws FileSystemException;
 
+    /**
+     * Returns an input stream for reading the file's content.
+     * <p>
+     * There may only be a single input or output stream open for the file at any time.
+     *
```
garydgregory: In Java 8 you must close tags, so a paragraph lives in a **`<p></p>`** block. IIRC, the Javadoc treats the first sentence of a Javadoc differently and it does not need to live in such a block.

boris-petrov: Done.

boris-petrov:
Javadoc comments fixed.

PR #60, opened 2019-04-26
VFS-609: VFS SFTP doesn't support a private key as byte array

RoTkay:
I really need feature with specifying private key in bytes. I guess that I'm not alone with this need.
Please, check my pull request. It's my first time making pull request in this project. If I made something wrong, I would be glad to hear your comments.
P.S. I took code changes from https://issues.apache.org/jira/browse/VFS-609 that stevezhuang attached to ticket, but that code wasn't placed to repo.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/IdentityInfo.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/IdentityInfo.java
@@ -26,6 +26,8 @@ import java.io.File;
 public class IdentityInfo {
     private final File privateKey;
     private final File publicKey;
+    private final byte[] privateKeyBytes;
+    private final byte[] publicKeyBytes;
     private final byte[] passPhrase;
 
     /**
@@ -69,6 +71,43 @@ public class IdentityInfo {
         this.privateKey = privateKey;
         this.publicKey = publicKey;
         this.passPhrase = passPhrase;
+        this.privateKeyBytes = null;
+        this.publicKeyBytes = null;
+    }
+
+
+    /**
+     * Constructs an identity info with private and public key and passphrase for the private key.
+     * <p>
+     *
+     * @param privateKey Private key bytes
+     * @param publicKey The public key part used for connections with exchange of certificates (can be {@code null})
+     * @param passPhrase The passphrase to decrypt the private key (can be {@code null} if no passphrase is used)
+     * @since 2.1
+     */
+    public IdentityInfo(final byte[] privateKey, final byte[] publicKey, final byte[] passPhrase)
+    {
+        this.privateKey = null;
+        this.publicKey = null;
+        this.privateKeyBytes = privateKey;
+        this.publicKeyBytes = publicKey;
+        this.passPhrase = passPhrase;
+    }
+
+    /**
+     * Constructs an identity info with private and passphrase for the private key.
+     * <p>
+     *
+     * @param privateKey Private key bytes
+     * @param passPhrase The passphrase to decrypt the private key (can be {@code null} if no passphrase is used)
+     * @since 2.1
+     */
+    public IdentityInfo(final byte[] privateKey, final byte[] passPhrase) {
+        this.privateKey = null;
+        this.publicKey = null;
+        this.publicKeyBytes = null;
+        this.privateKeyBytes = privateKey;
+        this.passPhrase = passPhrase;
     }
 
     /**
@@ -99,5 +138,25 @@ public class IdentityInfo {
      */
     public byte[] getPassPhrase() {
         return passPhrase;
+    }
+
+    /**
+     * Get private key bytes.
+     *
+     * @return private key
+     */
+    public byte[] getPrivateKeyBytes()
+    {
+        return privateKeyBytes;
+    }
+
+    /**
+     * Get public key bytes.
+     *
+     * @return public key
+     */
+    public byte[] getPublicKeyBytes()
+    {
+        return publicKeyBytes;
     }
 }
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpClientFactory.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpClientFactory.java
@@ -170,13 +170,13 @@ public final class SftpClientFactory {
             throws FileSystemException {
         if (identities != null) {
             for (final IdentityInfo info : identities) {
-                addIndentity(jsch, info);
+                addIdentity(jsch, info);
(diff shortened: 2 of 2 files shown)


commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/IdentityInfo.java
```diff
@@ -69,6 +71,43 @@ public IdentityInfo(final File privateKey, final File publicKey, final byte[] pa
...
         this.publicKey = publicKey;
         this.passPhrase = passPhrase;
+        this.privateKeyBytes = null;
+        this.publicKeyBytes = null;
+    }
+
+
+    /**
+     * Constructs an identity info with private and public key and passphrase for the private key.
+     * <p>
```
garydgregory: Open para tag is not needed.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/IdentityInfo.java
```diff
@@ -69,6 +71,43 @@ public IdentityInfo(final File privateKey, final File publicKey, final byte[] pa
...
+     *
+     * @param privateKey Private key bytes
+     * @param publicKey The public key part used for connections with exchange of certificates (can be {@code null})
+     * @param passPhrase The passphrase to decrypt the private key (can be {@code null} if no passphrase is used)
+     * @since 2.1
```
garydgregory: The next version will be 2.4, not 2.1.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/IdentityInfo.java
```diff
@@ -69,6 +71,43 @@ public IdentityInfo(final File privateKey, final File publicKey, final byte[] pa
...
+        this.privateKey = null;
+        this.publicKey = null;
+        this.privateKeyBytes = privateKey;
+        this.publicKeyBytes = publicKey;
+        this.passPhrase = passPhrase;
+    }
+
+    /**
+     * Constructs an identity info with private and passphrase for the private key.
+     * <p>
```
garydgregory: See above.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/IdentityInfo.java
```diff
@@ -69,6 +71,43 @@ public IdentityInfo(final File privateKey, final File publicKey, final byte[] pa
...
+     *
+     * @param privateKey Private key bytes
+     * @param passPhrase The passphrase to decrypt the private key (can be {@code null} if no passphrase is used)
+     * @since 2.1
```
garydgregory: See above.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/IdentityInfo.java
```diff
@@ -100,4 +139,24 @@ public File getPublicKey() {
     public byte[] getPassPhrase() {
         return passPhrase;
     }
+
+    /**
+     * Get private key bytes.
```
garydgregory: Javadoc should use "Gets..."

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/IdentityInfo.java
```diff
@@ -100,4 +139,24 @@ public File getPublicKey() {
...
+     *
+     * @return private key
+     */
+    public byte[] getPrivateKeyBytes()
+    {
+        return privateKeyBytes;
+    }
+
+    /**
+     * Get public key bytes.
```
garydgregory: See above.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpClientFactory.java
```diff
@@ -195,11 +195,17 @@ private static void setConfigRepository(final JSch jsch, final File sshDir, fina
         }
     }
 
-    private static void addIndentity(final JSch jsch, final IdentityInfo info) throws FileSystemException {
+    private static void addIdentity(final JSch jsch, final IdentityInfo info) throws FileSystemException {
```
garydgregory: Good catch! :-)

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpClientFactory.java
```diff
@@ -195,11 +195,17 @@ private static void setConfigRepository(final JSch jsch, final File sshDir, fina
         }
     }
 
-    private static void addIndentity(final JSch jsch, final IdentityInfo info) throws FileSystemException {
+    private static void addIdentity(final JSch jsch, final IdentityInfo info) throws FileSystemException {
         try {
-            final String privateKeyFile = info.getPrivateKey() != null ? info.getPrivateKey().getAbsolutePath() : null;
-            final String publicKeyFile = info.getPublicKey() != null ? info.getPublicKey().getAbsolutePath() : null;
-            jsch.addIdentity(privateKeyFile, publicKeyFile, info.getPassPhrase());
+            if (info.getPrivateKey() != null) {
```
garydgregory: It is confusing if more than one setting is used but one overrides the other silently. Wouldn't throwing an IllegalStateException be clearer?

RoTkay: Sorry, I didn't catch what do you mean by that. Throw` IllegalStateException` in which case?

garydgregory: What should happen if the `info` contains BOTH a private key file and private key bytes? Right now the file quietly overrides the bytes. This is good for compatibility. If neither are set you get an exception, which is new behavior. 
Should we throw an exception if both are set?
Is it possible for the provider to be used _without_ a PK?

garydgregory: Ah, I see. The issue I have now is that you have to know how which ctor you called to know which getters are valid, which is semi-lame. Let me see if I can adapt your code...

garydgregory:
Please see git master and the new class `org.apache.commons.vfs2.provider.sftp.BytesIdentityInfo`.

RoTkay:
Okey, saw your solution. Thanks for adding bytes approach! And thanks for your comments too :)

PR #61, opened 2019-04-29
Fix CI builds for Java 11

abashev:
Bugfix for broken TLS 1.3 in latest Java 11 https://bugs.openjdk.java.net/browse/JDK-8211806

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -248,7 +248,7 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
             <!-- to increase memory for tests on Travis CI -->
-            <argLine>-Xmx64m</argLine>
+            <argLine>-Xmx64m -Djdk.tls.client.protocols=TLSv1.2</argLine>
         </configuration>
       </plugin>
       <plugin>


garydgregory:
You need to redo this PR: there appear to be many noisy changes due to line-ending changes or whitespace tab vs. space changes, or I do not know what.

abashev:
@garydgregory done

garydgregory:
Thank you @abashev.

I took a slightly different approach by activating a profile on Java 11 and above with the TLS arg line.

This makes it clear IMO that we only need the hack (in the good sense of the word) on Java 11 and up.

Gary

abashev:
@garydgregory hm, ok, but travis builds should be green to get an understanding is everything alright. What do you think about this solution https://github.com/abashev/commons-vfs/commit/be31fce334fd2fbbd0a1953cfec2cc4e6cdc24b2

garydgregory:
It seems like this would apply to ALL Java setups, which I do not think is what we want. 

For Java EAs, we should either get it to work, or not test on EAs.

WDYT?

abashev:
https://github.com/apache/commons-vfs/pull/62

PR #62, opened 2019-04-29
Don’t fail on openjdk-ea builds

abashev:


garydgregory:
The problems seems like a Travis issue:
```
Installing openjdk-ea
$ export JAVA_HOME=~/openjdk-ea
$ export PATH="$JAVA_HOME/bin:$PATH"
$ ~/bin/install-jdk.sh --target "/home/travis/openjdk-ea" --workspace "/home/travis/.cache/install-jdk" --feature "ea" --license "GPL" --cacerts
install-jdk.sh 2019-04-18
Couldn't determine a download url for 13-GPL on linux-x64
The command "~/bin/install-jdk.sh --target "/home/travis/openjdk-ea" --workspace "/home/travis/.cache/install-jdk" --feature "ea" --license "GPL" --cacerts" failed and exited with 1 during .
Your build has been stopped.
```
Maybe we have to specify a specific OS setup like 'trusty' or some other version?

abashev:
@garydgregory yes, it is a Travis issue - there is no way to specify path or version, just `- openjdk-ea
` as the environment. So it is like well-known workaround - https://www.deps.co/guides/travis-ci-latest-java/ 'Ignoring failures with pre-release JDK’s'

garydgregory:
I just don't understand why we have other builds in Commons that can use the OpenJDK EA without any issue. In this case, we can't we even start to use it. So there must be some configuration thing happening that is allowing some of our Travis builds to work like https://travis-ci.org/apache/commons-text

abashev:
@garydgregory done, but builds will take more time in compare to docker environment

garydgregory:
Hi @abashev:

I am kinda -1 on openjdk-ea as allowing to fail since we know it always fails. What I was trying to say, unsuccessfully ;-) is either:
- Let's not even put openjdk-ea in the list of jdks if we then say it is allowed to fail, since it fails always
- Make openjdk-ea work, are least work enough that the build starts.

Thoughts?

Gary

abashev:
@garydgregory I'm not sure how do you mean.

image

EA build is up and running, I left it inside failed just in case. Is it not enough?

garydgregory:
I'd like to understand why the PR build passes:
https://travis-ci.org/apache/commons-vfs/builds/526143774
AND that build actually ran.

While the master build failed, _not even having started the build_:
https://travis-ci.org/apache/commons-vfs/jobs/526040009

abashev:
Because of this one https://github.com/apache/commons-vfs/pull/62/commits/44700e19bdb03e3d38c43f2b361fdb9661ccfc9a

abashev:
@garydgregory do you need more explanation about the fix or did I miss something before merging PR?

abashev:
@garydgregory ??

garydgregory:
Nothing to do IMO, all builds are green on https://travis-ci.org/apache/commons-vfs

PR #63, opened 2019-05-13
Use codecov.io for calculating code coverage

abashev:
I have no relations with codecov.io but I think it could be useful for this project (I used it for a while for vfs-s3 and it works fine). There are a few pros for it

1. It looks like coverall never worked for commons-vfs
image
1. It looks like codecov more active in compare to coverall - more twitter followers, latest tweets and blog posts from May, codecov was a part of Github marketplace initial submission etc.
1. Don't need separated maven plugin - it supports jacoco reports and could be installed as one-liner inside travis.yaml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@
 language: java
 sudo: false
 
+install: true
+
 jdk:
   - openjdk8
   - openjdk11
@@ -25,4 +27,5 @@ jdk:
   - oraclejdk11
 
 after_success:
-  - mvn -V -B -e clean cobertura:cobertura coveralls:report
+  - mvn -V -B -e jacoco:report
+  - bash <(curl -s https://codecov.io/bash)
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Apache Commons VFS Project
 ===================
 
 [![Build Status](https://travis-ci.org/apache/commons-vfs.svg)](https://travis-ci.org/apache/commons-vfs)
-[![Coverage Status](https://coveralls.io/repos/apache/commons-vfs/badge.svg)](https://coveralls.io/r/apache/commons-vfs)
+[![Coverage Status](https://codecov.io/gh/apache/commons-vfs/branch/master/graph/badge.svg)](https://codecov.io/gh/apache/commons-vfs)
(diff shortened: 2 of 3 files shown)


garydgregory:
We already use JaCoCo and/or Cobertura for code coverage on most Apache Commons projects for sites you can generate locally and also for our own sites, I'd rather not add yet a third option. Ideally, we can standardize one one. Then more recently, some projects use Travis and Coveralls. We do have not bothered to sync up all projects on a single coverage platform.

Also the square graph above looks useless to me :-( that or my brain does not connect in that way. Look at the HTML reports below for something more useful.

Cobertura Example: https://commons.apache.org/proper/commons-io/cobertura/index.html
JaCoCo Example: https://commons.apache.org/proper/commons-lang/jacoco/index.html

abashev:
It looks like a little misunderstanding about my changes - there are two components work together on code coverage. One is an agent for gathering coverage data (JaCoCo or Cobertura) and another is external service to gather historical data (Codecov or Coverall).

Cobertura is abandoned project, it doesn't support annotations and latest Java 8+ features. So I think we don't needto spend time on keeping this plugin inside pom. And JaCoCo is only one option that works fine with latest Java and has been supported by all code coverage services. I didn't put JaCoCo into <reporting> tag because I think single report doesn't make any sense and gives zero information about the project itself.

Because from my understanding the only thing that matter for code coverage is changing of coverage. And here come external services that gather historical data - Codecov and Coverall. If you are submitting huge pile of code then only changing of code coverage could say you did you spend enough efforts on writing tests or not. There is how it looks like for vfs-s3 https://codecov.io/gh/abashev/vfs-s3/commits

In first I wanted to keep Coverall and migrate vfs-s3 on it too. But it looks like coverall never worked for commons-vfs and you don't have any historical data about the coverage. Another issue is if you want to have coverall reports then need to add one more maven plugin into pom. In compare Codecov works directly with JaCoCo reports and don't need to update pom at all. And last but not least, Coverall looks neglected for me - latest tweet from 2017 https://twitter.com/CoverallsApp no any activity in issue tracker for months. And in compare Codecov is more active with more followers https://twitter.com/codecov?lang=en

You are main contributor so that is up to you - I can move JaCoCo into <reporting> and fix coverall integration but you should have in mind that there is an alternative.

PR #64, opened 2019-05-16
Fix AbstractFileName.getURI returning unencoded #-sign

boris-petrov:
We hit the same issue as [VFS-325](https://issues.apache.org/jira/projects/VFS/issues/VFS-325) but not for local files but rather for Samba. Extracting the logic from `LocalFileName` to the base `AbstractFileName` fixes it for any provider.


boris-petrov:
Thanks for merging this! Do you mind creating a new snapshot with the latest changes?

garydgregory:
Done.

PR #65, opened 2019-06-16
VFS-619 Fixes performance issue with SftpFileObject.getChildren()

DavidSeptimus:
This PR removes an unnecessary invocation of `SftpObject.getType()` by `refresh()`. 

This invocation causes performance degradation of `SftpFileObject.getChildren()` when `CacheStrategy.ON_RESOLVE` is in use as this forces a call to `refresh()`, which in turn calls `getType()`, making an additional `stat()` call to the SFTP server for each child.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileObject.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileObject.java
@@ -68,23 +68,6 @@ public class SftpFileObject extends AbstractFileObject<SftpFileSystem> {
     }
 
     /**
-     * @throws FileSystemException if error occurs.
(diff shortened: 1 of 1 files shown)


ecki:
It is probably better to remove this refresh implementation completely, the super implementation will call detach/doDetach which already sets `attrs = null`

DavidSeptimus:
That's a fair point.

garydgregory:
Squashing and merged. @DavidSeptimus & @ecki Thank you.

DavidSeptimus:
Ah, I just realized that I referenced the wrong Jira story. It should have been VFS-698.

https://issues.apache.org/jira/browse/VFS-698

garydgregory:
Fixed in `changes.xml`, thanks @DavidSeptimus.

PR #30, opened 2018-02-17
VFS-614 MonitorInputStream should not close the stream in "read"

ottobackwards:
MonitorInputStream exposes that it supports marking by way of it's super BufferedInputStream.  
Because it closes itself after a read which encounters the end of the stream however ( which is contrary to BufferedInputStream ) it breaks marking functionality.

This PR removes to the close() call on read after EOF, making the MonitorInputStream behave as BufferedInputStream, and as one would expect a Stream that supports marking would.  Tests are also added.

PR #67, opened 2019-06-18
MonitorInputStream should not close the stream in "read"

boris-petrov:
Similar to https://github.com/apache/commons-vfs/pull/30. That PR fixed only the `read` overload which accepts a buffer but missed the one with no arguments. This PR fixes also the other case.

See https://issues.apache.org/jira/projects/VFS/issues/VFS-614.


ottobackwards:
+1, sorry to have missed this in my PR

PR #69, opened 2019-08-26
VFS-727: Replaced VFS.getManager() calls to more local FileSystemManager access.

elmuerte:
All except the call in org.apache.commons.vfs2.provider.HostFileNameParser have been replaced to locally reachable instances of FileSystemManager.
I have a change for HostFileNameParser, but it breaks the test cases which call parseUri without a context.


ecki:
I agree that we should not use the static getManager method at all, and it is probably best to deal with the failed tests in a separate change (i would suggest the test constructs a manner and hands it over and we change the teste function to check for null and fallback to getManager in worst case.

BTW, i also have a Change which does require less of that scheme prefix checking, maybe all of that can be removed as well.

PR #72, opened 2019-10-09
VFS-299 Add failing tests for missing filesystem listener removal in DefaultFileMonitor

f4lco:


garydgregory:
I merged a slightly different version:
- Use `@Ignore` to pass the build.
- Renamed the class to match other test classes in the package.

Now we need a fix...

f4lco:
Yes, thank you @garydgregory!

PR #73, opened 2019-10-11
DefaultFileMonitor addFile() no longer works with directories since VFS-702

f4lco:
I currently assume the change from VFS-702 ([diff](https://github.com/apache/commons-vfs/commit/cb82adc901f571e7124e400fb60886851c941732#diff-33e19a4a4f0966e0d3b1d80f695a2f87)) misses one important case. Previously, if directories were added to non-recursive file monitors, their direct file children have been registered for watching. Currently, that operation is a no-op.

The currently failing test case included in this PR demonstrates this behavior. The same test case is passing before the refactoring of `addFile` as shown in the branch [filemonitor-before-addfile-refactoring](https://github.com/f4lco/commons-vfs/commits/filemonitor-before-addfile-refactoring) in my fork.

### DIFF
--- a/commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/test/DefaultFileMonitorTest.java
+++ b/commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/test/DefaultFileMonitorTest.java
@@ -216,6 +216,42 @@ public class DefaultFileMonitorTest extends AbstractVfsTestCase {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


garydgregory:
@f4lco 
It looks like the commit you refer also fixed a bug which it seems you depended on: If you want to monitor a directory, you need to enable recursion with `setRecursive(true)` before calling `addFile(FileObject)`.

Note: I updated `CountingListener` to could all three API calls.

f4lco:
Yeah, I think the main reason for the PR and test case has been 'did not work as before'. So the behavior has changed for the better / another bug has been fixed.  That's totally fine, thank you for the feedback. Good job 👍

PR #74, opened 2019-10-23
VFS-687: adding http5 and http5s providers

woonsan:


commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/http5/Http5FileContentInfoFactory.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.commons.vfs2.impl.DefaultFileContentInfo;
+import org.apache.commons.vfs2.util.FileObjectUtils;
+import org.apache.hc.core5.http.ContentType;
+import org.apache.hc.core5.http.Header;
+import org.apache.hc.core5.http.HttpHeaders;
+import org.apache.hc.core5.http.HttpResponse;
+
+/**
+ * Creates <code>FileContentInfoFactory</code> instances for http5 provider.
+ */
```
garydgregory: New types should use `@since `

michael-o:
The name is confusing. Shouldn't this be hc5 and hc5s at most?

garydgregory:
Would you rename version 4 to hc4 as well?

michael-o:
Yes, people could simply confuse with HTTP, version 4 or 5.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/http5/Http5FileContentInfoFactory.java
woonsan: Done. Thanks!

woonsan:
I have assumed that Commons VFS would implement http/s providers using HC libraries. Unless we want to use other libraries than HC, http3, http4 and http5 do not seem confusing to me. At the moment, 'http3' is aliased as 'http', but we may deprecate that and give 'http' alias to http4 someday, and so forth.

garydgregory:
> I have assumed that Commons VFS would implement http/s providers using HC libraries. Unless we want to use other libraries than HC, http3, http4 and http5 do not seem confusing to me. At the moment, 'http3' is aliased as 'http', but we may deprecate that and give 'http' alias to http4 someday, and so forth.

I'm OK with our current scheme. Unless someone proposes something better but it seems that hc4 and hc5 are a little too odd. Sure HTTP/2 is released spec. But until HTTP/3 ,HTTP/4 and HTTP/5 are all out there, I'm not sure there is that much room for confusion. I'd like to bring this in to make future PRs of smaller scale.

PR #75, opened 2019-12-16
VFS-590 VFS-617 Proceed with moveTo operation if the exec channel for…

apearc03:
Pull request based on previous pr27 here.
https://github.com/apache/commons-vfs/pull/27

I've restructured the org.apache.commons.vfs2.provider.sftp.test package so that it contains an abstract test case similar to org.apache.commons.vfs2.provider.ftps.test.
The existing SFTP test cases are now split into individual implementations.

I've added a unit test to replicate the exec channel being closed for permission checks. An IOException is thrown during permission checks "id -G" or "id -u" meaning SftpFileSystem#executeCommand returns a failed exit status code.


commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/sftp/test/SftpProviderClosedExecChannelTestCase.java
```diff
@@ -0,0 +1,22 @@
+package org.apache.commons.vfs2.provider.sftp.test;
```
garydgregory: ASL header is missing causing the build to fail on Travis.

commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/sftp/test/SftpProviderStreamProxyModeTestCase.java
```diff
@@ -0,0 +1,88 @@
+package org.apache.commons.vfs2.provider.sftp.test;
+
```
garydgregory: ASL header is missing causing the build to fail on Travis.

garydgregory:
@apearc03 
Thank you for your PR.
Please rebase on master and add the missing Apache License headers.
Thank you!

commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/sftp/test/SftpProviderStreamProxyModeTestCase.java
apearc03: added

commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/sftp/test/SftpProviderClosedExecChannelTestCase.java
apearc03: added

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
```diff
@@ -76,12 +80,18 @@
...
+     */
+    private boolean execDisabled;
+
     protected SftpFileSystem(final GenericFileName rootName, final Session session,
             final FileSystemOptions fileSystemOptions) {
         super(rootName, null, fileSystemOptions);
         this.session = Objects.requireNonNull(session, "session");
         this.connectTimeoutMillis = SftpFileSystemConfigBuilder.getInstance()
                 .getConnectTimeoutMillis(fileSystemOptions);
+        detectExecDisabled();
```
garydgregory: @apearc03 

Thank you for updating the ASL headers.

Unless you expect the state of a server to change and re-detect `execDisabled`, please make `execDisabled` `final` and make `detectExecDisabled()` return a `boolean`.  A boolean method would be cleaner anyway IMO. Then the ctor can say 'this.execDisable = detectExecDisabled();`

We can always drop the `final` on the ivar later if `detectExecDisabled()` needs to be called post construction.

apearc03: Made changes to cleanup exec ```execDisabled``` initialization.

LoicDelorme:
@garydgregory any new release incoming which includes this fix?

garydgregory:
> @garydgregory any new release incoming which includes this fix?

I might take the time over the holidays (b/w now and the new year) to cut some Commons releases... no guarantees. There are a few Commons components that I can see releasing aside from this one as well. You can subscribe to our Commons Developer's mailing list to track progress, https://commons.apache.org/mail-lists.html

simonalexander2005:
Hi, did this go anywhere? I'm having the same issue in 2.9.0

garydgregory:
Hi @simonalexander2005 
The PR was mered, so you should try git master or a snapshot build from https://repository.apache.org/content/repositories/snapshots/

PR #76, opened 2019-12-26
[commons-vfs] clean up tests and simplify assertions

grimreaper:
This is a mostly mechanical change to tests such that it is easier to
convert from junit 3 to junit 4

### DIFF
--- a/commons-vfs2/src/test/java/org/apache/commons/AbstractVfsTestCase.java
+++ b/commons-vfs2/src/test/java/org/apache/commons/AbstractVfsTestCase.java
@@ -18,7 +18,6 @@ package org.apache.commons;
 
 import java.io.File;
 import java.io.IOException;
-import java.lang.reflect.Method;
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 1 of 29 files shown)


commons-vfs2/src/test/java/org/apache/commons/AbstractVfsTestCase.java
```diff
@@ -136,7 +118,7 @@ public static File getTestDirectory(final String name) {
     /**
      * Makes a file canonical
      */
-    public static File getCanonicalFile(final File file) {
+    private static File getCanonicalFile(final File file) {
```
garydgregory: There is no need to hide methods in tests, it just makes reusing code more difficult when using discovery features in IDEs like CTRL-space for method names.

grimreaper: I find it makes it easier to detect unused code along with all the other typical reasons to default to private. That said I don't feel strongly about this so I'll just revert.

PR #78, opened 2020-01-19
fix bug for class not found:org.apache.commons.vfs2.provider.webdav.WebdavFileProvider

XenoAmess:
1.fix bug for class not found:org.apache.commons.vfs2.provider.webdav.WebdavFileProvider since 2.5.0

this bug is caused by the remove of WebdavFileProvider, and when try to load WebdavFileProvider if the user does not have class WebdavFileProvider, then explode.

now the code will check the class as well as its dependency when load.(if the class itself does not exist then return)

2.add clean room test
This bug happened and even with this bug the tests now can pass.
that is because you include all things needed in test scope guys.
and we users do not usually include them all.
so when something went wrong and vfs can only run correctly when user code imports some repo(like this time), the current test mechanism can never tell.
So I add a clean room test for it.

### DIFF
--- a/commons-vfs2-cleanroomtest/pom.xml
+++ b/commons-vfs2-cleanroomtest/pom.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
+   The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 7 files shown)


XenoAmess:
@garydgregory ping?

woonsan:
Hi @XenoAmess ,
(CC: @garydgregory )

Thanks again for reporting and helping with problem determination!
I've filed a JIRA ticket with a simple improvement in a new PR ( https://github.com/apache/commons-vfs/pull/84 ) to avoid this kind of corner case in the future:
- https://issues.apache.org/jira/browse/VFS-759

Cheers,

Woonsan

PR #79, opened 2020-02-05
add possibility to configure if exec for sftp is disabled

ddg-igh:
One of our customer has an interesting sftp server. 
The server does not allow any further channels to be established after the execution of the exec command (getUid()).
So it is not possible to do any filesystem operations, because at first the exec command is executed to test if exec is disabled.
Maybe the server things we try to break the security or something else.

So my idea to solve this problem is to make execDisabled configurable. This is done here.


garydgregory:
@ddg-igh 
-1 You need to resubmit this PR as _every single line_ in SftpFileSystem.java is marked as changed.

PR #80, opened 2020-02-05
[SFTP] Configure whether exec detection is enabled

ddg-igh:
One of our customer has an interesting sftp server.
The server does not allow any further channels to be established after the execution of the exec command (getUid()).
So it is not possible to do any filesystem operations, because at first the exec command is executed to test if exec is disabled.
Maybe the server things we try to break the security or something else.

So my idea to solve this problem is to make execDisabled configurable. This is done here.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
@@ -91,7 +91,13 @@ public class SftpFileSystem extends AbstractFileSystem {
         this.session = Objects.requireNonNull(session, "session");
         this.connectTimeoutMillis = SftpFileSystemConfigBuilder.getInstance()
                 .getConnectTimeoutMillis(fileSystemOptions);
-        this.execDisabled = detectExecDisabled();
+        
+        if (SftpFileSystemConfigBuilder.getInstance().isDisableDetectExecChannel(fileSystemOptions)){
+            this.execDisabled = true;
+        }
+        else {
+            this.execDisabled = detectExecDisabled();
+        }
     }
 
     @Override
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
@@ -92,10 +92,9 @@ public class SftpFileSystem extends AbstractFileSystem {
         this.connectTimeoutMillis = SftpFileSystemConfigBuilder.getInstance()
                 .getConnectTimeoutMillis(fileSystemOptions);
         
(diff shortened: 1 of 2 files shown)


commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
```diff
@@ -91,7 +91,13 @@ protected SftpFileSystem(final GenericFileName rootName, final Session session,
         this.session = Objects.requireNonNull(session, "session");
         this.connectTimeoutMillis = SftpFileSystemConfigBuilder.getInstance()
                 .getConnectTimeoutMillis(fileSystemOptions);
-        this.execDisabled = detectExecDisabled();
+        
+        if (SftpFileSystemConfigBuilder.getInstance().isExecEnabled(fileSystemOptions)){
+            this.execDisabled = detectExecDisabled();
+        }
```
garydgregory: formatting should be "} else {"

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystemConfigBuilder.java
```diff
@@ -268,6 +269,21 @@ public ConfigRepository getConfigRepository(final FileSystemOptions opts) {
     public boolean isLoadOpenSSHConfig(final FileSystemOptions opts) {
         return this.getBoolean(opts, LOAD_OPENSSH_CONFIG, Boolean.FALSE);
     }
+    
+    /**
+     * Returns {@link Boolean#TRUE} if exec should be tested and enabled.
```
garydgregory: The API returns a `boolean`, not a `Boolean`, so the Javadoc should refer to `true`, not `Boolean#TRUE}.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
```diff
@@ -91,7 +91,13 @@ protected SftpFileSystem(final GenericFileName rootName, final Session session,
         this.session = Objects.requireNonNull(session, "session");
         this.connectTimeoutMillis = SftpFileSystemConfigBuilder.getInstance()
                 .getConnectTimeoutMillis(fileSystemOptions);
-        this.execDisabled = detectExecDisabled();
+        
+        if (SftpFileSystemConfigBuilder.getInstance().isExecEnabled(fileSystemOptions)){
+            this.execDisabled = detectExecDisabled();
```
garydgregory: This feature disables _detection_ of the exec channel, so it should be called as such IMO: `EXEC_ENABLED` -> `DISABLE_DETECT_EXEC_CHANEL`.

garydgregory:
@ddg-igh ,

Thank you for your PR changes. Please see my comments. Also, please rebase on master, I just fixed an issue in this provider.

ottobackwards:
+1 to a configuration only fix, although I would want to understand if there is better way to execute the original intent without more configuration nobs

PR #81, opened 2020-02-13
bugfix webdav4s request urls

ddg-igh:
When using webdav4s the requests will be send as "http" and not "https".
So this will always fail when the webdav server does not allow "http" requests.

### DIFF
--- a/commons-vfs2-jackrabbit2/src/main/java/org/apache/commons/vfs2/provider/webdav4s/Webdav4sFileSystem.java
+++ b/commons-vfs2-jackrabbit2/src/main/java/org/apache/commons/vfs2/provider/webdav4s/Webdav4sFileSystem.java
@@ -14,10 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.commons.vfs2.provider.webdav4;
-
-import java.net.URLStreamHandler;
-import java.util.Collection;
+package org.apache.commons.vfs2.provider.webdav4s;
 
 import org.apache.commons.vfs2.Capability;
 import org.apache.commons.vfs2.FileName;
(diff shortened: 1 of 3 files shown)


commons-vfs2-jackrabbit2/src/main/java/org/apache/commons/vfs2/provider/webdav4s/Webdav4sFileSystem.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.http.client.HttpClient;
+import org.apache.http.client.protocol.HttpClientContext;
+
+import java.net.URLStreamHandler;
+import java.util.Collection;
+
+/**
+ * A WebDAV file system based on HTTP4.
+ *
+ * @since 2.5.0
```
garydgregory: The next version will be 2.7.0.

commons-vfs2-jackrabbit2/src/main/java/org/apache/commons/vfs2/provider/webdav4s/Webdav4sFileSystem.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Creates a file object. This method is called only if the requested file is not cached.
+     *
+     * @param name the FileName.
+     * @return The created FileObject.
+     */
+    @Override
+    protected FileObject createFile(final AbstractFileName name) throws Exception {
+        return new Webdav4FileObject(name, this,true){};
```
garydgregory: Missing space after the ',' but it would see cleaner to create a Webdav4FileObject subclass.
Waiting on @kinow to opine.

woonsan:
Hi @ddg-igh ,
(CC: @garydgregory )

I've just copied your report to https://issues.apache.org/jira/browse/VFS-758. Thank you very much for reporting! It was my mistake not to test this basic thing. Sorry about that.
I'd like to look into the issue this weekend and let you guys know.

Regards,

Woonsan

woonsan:
Hi @ddg-igh ,
(CC: @garydgregory )

I think I found a simpler fix: https://github.com/apache/commons-vfs/pull/83
Basically, all Http4FileObject and its descendants have internal URIs, converted from the logical one to physical one already. So, it seems simpler just to use `#getInternalURI().getScheme()` in `Webdav4FileObject.toUrlString(GenericURLFileName, boolean)`.

Thanks again!

Woonsan

ddg-igh:
Hi @woonsan,
this looks like a nice solution. So i can close this request. 
The only thing someone should document is that the Webdav4sFileProvider should only be bound to schemes with an "s" at the end of the name. In case that people will manually bind the provider to schemes. Otherwise this will not work.

PR #83, opened 2020-02-16
VFS-758: resolve url scheme by its internaly-resolved uri scheme

woonsan:
Hi @garydgregory ,
I think this will resolve the VFS-758, which is basically caused by the hard-coded "http" in Webdav4FileObject#toUrlString(GenericURLFileName, boolean), by which some webdav specific http invocation is done by resolving "internal" http(s) urls.
Regards,
Woonsan


woonsan:
Closing this as it's to be done in https://github.com/apache/commons-vfs/pull/143.

PR #87, opened 2020-02-22
Fix travis-ci to make allow_failures: - openjdk-ea due to https://jira.apache.org/jira/browse/VFS-765

XenoAmess:
this project cannot pass travis since Jul 22, 2019.
If you do care about the error on openjdk-ea, you shall fix it.
If you do not care, then shall add openjdk-ea to allow_failures.
It is not recommended to keep project failing travis for such a long time (several months till today.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,12 @@ jdk:
   - openjdk11
   - openjdk12
   - openjdk13
-  - openjdk-ea
   - oraclejdk11
+  - openjdk-ea
+
+matrix:
+  allow_failures:
+    - jdk: openjdk-ea
 
 after_success:
   - mvn -V -B -e clean cobertura:cobertura coveralls:report


ecki:
Thanks, will merge it when the builds succeed.

BTW if you notice such a thing, don't complain about month of people not noticing, just open a bug, this also makes changelog nicer.

I suspect this is a dependency problem, have your actually checked if it can be fixed, instead?

ecki:
> I suspect this is a dependency problem, have your actually checked if it can be fixed, instead?

Hm ok, is a WebDAV Test failure. I created https://jira.apache.org/jira/browse/VFS-765 - maybe you can add this to your commit message?

XenoAmess:
> BTW if you notice such a thing, don't complain about month of people not noticing, just open a bug, this also makes changelog nicer.

sorry for being rude.

> I suspect this is a dependency problem, have your actually checked if it can be fixed, instead?

Nope I can't.
I use jdk-15 ea on windows to run this project, and cannot get this error.
In other words jackrabit1 can pass build on my PC.

PR #88, opened 2020-05-19
Fix for failed test cases when running with root

PeterAlfredLee:
Some tests are failed when running with root. This is a fix for them.

### DIFF
--- a/commons-vfs2/src/test/java/org/apache/commons/vfs2/filter/CanReadFileFilterTest.java
+++ b/commons-vfs2/src/test/java/org/apache/commons/vfs2/filter/CanReadFileFilterTest.java
@@ -24,6 +24,7 @@ import org.apache.commons.vfs2.FileFilterSelector;
 import org.apache.commons.vfs2.FileObject;
 import org.apache.commons.vfs2.FileSelectInfo;
 import org.apache.commons.vfs2.FileSystemException;
+import org.apache.commons.vfs2.util.Os;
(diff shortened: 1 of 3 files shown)


garydgregory:
-1 This is way to restrictive and makes no sense to me: Only non-root users can perform certain tests? The root user should all all rights over all files.

PeterAlfredLee:
> Only non-root users can perform certain tests?

Nope. But these assert would fail if you are running the tests using root.
I'm running these tests on Ubuntu 18.04. The tests are failing if I'm running with root, and they are passed with any non-root user. And the tests are passed on Windows.
I checked the file after calling `file.setWritable(false)`, and it shows the file is not writable(-r--r--r--) when I run `ls -l`. At the same time I'm getting true when calling `file.canWrite()`. Seems the root user can bypass the permission and can write to files that are readonly for root.
See also : [File.canWrite and Files.isWritable not giving correct value on Linux](https://stackoverflow.com/questions/28366433/file-canwrite-and-files-iswritable-not-giving-correct-value-on-linux)

garydgregory:
Hi @PeterAlfredLee 
If this is still a concern, may you please rebase on master?

garydgregory:
Hi @PeterAlfredLee
If this is still a concern, may you please rebase on master?

garydgregory:
Ping?

garydgregory:
Ping?

garydgregory:
Closing, no reply in months.

PR #89, opened 2020-05-24
[VFS-443] added function FileSystemManager.toFile(FileObject fileObject)

XenoAmess:
added function FileSystemManager.toFile(FileObject fileObject)
https://issues.apache.org/jira/browse/VFS-443

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/FileSystemManager.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/FileSystemManager.java
@@ -351,4 +351,12 @@ public interface FileSystemManager extends AutoCloseable {
      */
     FileObject toFileObject(File file) throws FileSystemException;
 
+    /**
+     * Converts {@link FileObject} to a local file.
+     *
+     * @param fileObject The file object to convert.
+     * @return the local file.
+     * @throws FileSystemException On error converting the file object.
+     */
+    File toFile(FileObject fileObject) throws FileSystemException;
 }
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/impl/DefaultFileSystemManager.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/impl/DefaultFileSystemManager.java
@@ -882,6 +882,27 @@ public class DefaultFileSystemManager implements FileSystemManager {
     }
 
     /**
+     * Converts {@link FileObject} to a local file.
+     *
+     * @param fileObject The file object to convert.
+     * @return the local file.
+     * @throws FileSystemException On error converting the file object.
+     */
+    @Override
+    public File toFile(FileObject fileObject) throws FileSystemException{
+        if(fileObject == null || !"file".equals(fileObject.getURL().getProtocol())){
+            return null;
+        }
+        File result = null;
+        try {
+            result = new File(fileObject.getName().getPathDecoded());
+        } catch (Exception e) {
+            throw new FileSystemException(e);
+        }
+        return result;
+    }
+
+    /**
      * Creates a layered file system.
      *
      * @param scheme The scheme to use.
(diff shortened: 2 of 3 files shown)


dwalluck:
Is it possible to get a local file for a remote object like an HTTP URL? Even if it must copy the input stream first, it would be nice to have such functionality provided.

XenoAmess:
> Is it possible to get a local file for a remote object like an HTTP URL? Even if it must copy the input stream first, it would be nice to have such functionality provided.

@dwalluck If you mean creating a temp file and clone the content into it, this can be done.
But should in another function.

dwalluck:
> @dwalluck If you mean creating a temp file and clone the content into it, that can be done.
> But should in another function.
> Whenever you create a temp file you should consider about when to delete it.

I am not sure if I will make sense, but I will try. I think there are two issues:

1. If I want to get a `java.io.File` from a `FileObject`, it isn't clear to me if this would be done automatically for a remote file (non-`file://` URL)
2. If I want to take advantage of the file cache, say for multiple calls to the same URL, then I think I want the cache to persist.

garydgregory:
-1 This is mixing two kinds of file systems which I do not think is a pattern we want to get into. Plus it breaks binary compatibility anyway.

XenoAmess:
> -1 This is mixing two kinds of file systems which I do not think is a pattern we want to get into. 

Don't think so, as we have a toFileObject function that convert a File to a FileObject.
It is just a, mirror side of toFileObject.

Besides, this is needed mainly for Swing and AWT, which MUST accept a File...
For example, a JFileChooser. (that is where I meet this demand)
I just cannot get a better idea about how to achieve it.

> Plus it breaks binary compatibility anyway.

Yes.
Maybe we can delay this pr to next Major release?

rmannibucau:
Thinking out loud: why not adding in local filesystem a FileOperation "ToLocalFile", will be backward compatible and provide this feature in a more natural way IMO (caller can test the ToLocalFile.class exists and call it if so, else fallback on downloading the file).

garydgregory:
See also @dwalluck's https://github.com/apache/commons-vfs/pull/89#issuecomment-647840475

pdinc-oss:
> Thinking out loud: why not adding in local filesystem a FileOperation "ToLocalFile", will be backward compatible and provide this feature in a more natural way IMO (caller can test the ToLocalFile.class exists and call it if so, else fallback on downloading the file).

While (I think) I understand the OPs intention, the implementation is way off as the PR comments react to.

If a File object is desired, it would have to be a subclass that wraps all of the capability of the FileObject, and MUST not assume any specifics (e.g. "file".equals...).

The File API is not that hard to implement, it even has well defined alternate values for most of the functions if it cannot be implemented (e.g. public long lastModified() Returns: ... 0L if the file does not exist).

rmannibucau:
@pdinc-oss well it would still breaks most of the API and keep in mind File API includes Path API now (so FilesystemProvider and so on). If I got it right, the goal of this jira is to get access to a File to match the UI requirement so instead of reimplementing a java filesystem OR loosing vfs features, I think adding an operation is not a that bad compromise since it fulfills the requested feature for all protocols in ~10 LoC compared to breaking the compatibility, hiding or adding a file cache (which is rarely a good idea in this kind of lib IMHO) or dopping vfs to use java FS instead. To have hacked something around that for a web app (overriding local FS to expose getLocalFile()), it does not sounds crazy to have a specific metadata extractor (through op or not, op just enables to not break anything) since you always hit some abstraction limitations. Another option is to add an unwrap(Class) method but thought operations were already 1-1 in terms of API.

XenoAmess:
@pdinc-oss
> The File API is not that hard to implement。

What is File API?
As far as I can see, File is a class, not an interface.
So if you mean we extend class File and override all functions it have?
So what if they add more functions in class File in future JDK releases?
Can we always override them quick and in time？

XenoAmess:
@pdinc-oss
besides it might NOT be as easy as you think...
for example, File have a function `public Path toPath()`
means we must implement a Path class as well.
(however Path be an interface, that is far better to implement it than File.)
Anyway, if we really should do it that way, I want to hear about others opinions.
I don't want to make it first and get refused later, because that will be a large change, and time consuming.
@garydgregory what's your ideas, gary? should we start such a try?

garydgregory:
I think we should close this in favor of using `FileObject.getPath().toFile()` instead.

XenoAmess:
> I think we should close this in favor of using `FileObject.getPath().toFile()` instead.

Good to hear there finally be a way to do so.
I will test it when I get home today.
Wait for several hours pls.

PR #92, opened 2020-06-10
[VFS-770] Gz createFileSystem fails

PeterAlfredLee:
The `.gz` is parsed as `application/octet-stream` by mime type and could not be successfully handled.
This PR is a fix for this.


commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/test/DefaultFileSystemManagerTest.java
```diff
@@ -63,4 +68,22 @@ public void testResolveFileObjectNullAbsolute() throws FileSystemException {
...
+
+
+    @Test
+    public void testCreateGzipFileSystem() throws IOException {
+        // setup .gz file for test
+        final File gzFile = new File("src/test/resources/test-data/test.gz");
+        final File newGzFile = File.createTempFile(getClass().getSimpleName(), ".gz");
+        newGzFile.deleteOnExit();
+        FileUtils.copyFile(gzFile, newGzFile);
+        FileSystemManager manager = VFS.getManager();
```
garydgregory, line 80: May you please either add a comment as to why this test needs its own copy of the file, or simplify the test. Can't the test use the file fixture as it stands? Assuming the test releasing resources works and there are no bugs at that level of course.

PeterAlfredLee: Sure. Will push again avoiding copying the file.

commons-vfs2/src/main/java/org/apache/commons/vfs2/impl/FileTypeMap.java
```diff
@@ -57,7 +58,7 @@ public String getScheme(final FileObject file) throws FileSystemException {
         // Check the file's mime type for a match
         final FileContent content = file.getContent();
         final String mimeType = content.getContentInfo().getContentType();
-        if (mimeType != null) {
+        if (mimeType != null && !mimeType.equals(MIME_TYPE_OCTET_STREAM)) {
```
garydgregory, line 61: This does not feel right. One the one hand we tell caller to call `addMimeType(String, String)` but then we ignore that entry for one specific value. That does not make sense. I'd like to hear from others who might be more familiar with this part of the code.

PeterAlfredLee: Not familiar with `mimeType` eithor, seems it's used in Internet media.

According to the [mimeType table](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) the `.gz` should be regarded as `application/gzip`, but we got a `application/octet-stream` here. Actually I diged a little here and found it's caused in `sun.net.www.MimeTable` and will treat extensions `.saveme, .dump, .hqx, .arc, .obj, .lib, .bin, .exe, .gz` as `application/octet-stream`.Not sure if this is a potential problem of `sun.net.www.MimeTable` or not. 

This is just a workaround for it.

PeterAlfredLee: Actually `application/octet-stream` is the default value for extensions not in the [mime type table](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types). I'm just adding one more extension check using `VFS's extensionMap` if we encountered a `application/octet-stream` here.

garydgregory:
Hi @PeterAlfredLee 
May you please rebase on master?

garydgregory:
https://issues.apache.org/jira/browse/VFS-770

garydgregory:
Implemented differently in git master.

PR #93, opened 2020-06-15
[VFS-624] Fix for read() in constructors of LocalFileRandomAccessContent and RamFileRandomAccessContent

PeterAlfredLee:
As [VFS-624](https://issues.apache.org/jira/projects/VFS/issues/VFS-624?filter=allopenissues) described, the `read`of the inner class in `LocalFileRandomAccessContent` would return `-1` if it's reading a `0xFF` in file.
This is a fix for it.

And the same problem also exists in `RamFileRandomAccessContent` and the fix is also included in this PR.


commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/local/LocalFileRandomAccessContent.java
```diff
@@ -45,7 +45,7 @@
                 @Override
                 public int read() throws IOException {
                     try {
-                        return raf.readByte();
+                        return raf.readByte() & 0xFF;
```
garydgregory, line 48: Let's avoid repeated use of a magic number. Please refactor into a constant that reused in the code and tests. Same for -1.

PeterAlfredLee: Sure.
Seems we need a place to store the constant `0xFF`. IMO we should have a class to store constants like `0xFF`, which seems is not presented in VFS. What about adding a `VFSConstants` class in `org.apache.commons.vfs2.util`?

garydgregory: For now I would just add it as a constant in `RamFileRandomAccessContent`. The only other place I see it used, is once in an unrelated sandbox class.

PR #95, opened 2020-06-15
[VFS-664] Fix for file names with exclamation mark can not be read

PeterAlfredLee:
This is a fix for problems that file names with exclamation mark in zip file can not be successfully be read.
And this problem also exists in tar, tgz and tbz2. The fix for them are also included in this PR.


PeterAlfredLee:
And it seems the [VFS-225](https://issues.apache.org/jira/projects/VFS/issues/VFS-225?filter=allopenissues) is a duplicate issue with this one.

PR #96, opened 2020-06-28
Modify some loop using stream API

PeterAlfredLee:
Modify some loop using stream API.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/AbstractVfsContainer.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/AbstractVfsContainer.java
@@ -17,6 +17,7 @@
 package org.apache.commons.vfs2.provider;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 
 import org.apache.commons.vfs2.FileSystemException;
 
@@ -81,11 +82,7 @@ public abstract class AbstractVfsContainer extends AbstractVfsComponent {
         }
 
         // Close all components
-        for (final Object component : toclose) {
-            if (component instanceof VfsComponent) {
-                final VfsComponent vfsComponent = (VfsComponent) component;
-                vfsComponent.close();
-            }
-        }
+        Arrays.stream(toclose).filter(component -> component instanceof VfsComponent)
+                              .forEach(component -> ((VfsComponent) component).close());
     }
 }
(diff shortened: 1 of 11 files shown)


commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/AbstractVfsContainer.java
```diff
@@ -81,11 +82,7 @@ public void close() {
...
 
         // Close all components
-        for (final Object component : toclose) {
-            if (component instanceof VfsComponent) {
-                final VfsComponent vfsComponent = (VfsComponent) component;
-                vfsComponent.close();
-            }
-        }
+        Arrays.stream(toclose).filter(component -> component instanceof VfsComponent)
+                              .forEach(component -> ((VfsComponent) component).close());
```
garydgregory, line 86: @PeterAlfredLee 
May you please rebase on `master`, test and cast to `java.io.Closeable` instead of `VfsComponent`.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/CompositeFileProvider.java
```diff
@@ -53,12 +55,7 @@ public FileObject findFile(final FileObject baseFile, final String uri, final Fi
         final StringBuilder buf = new StringBuilder(INITIAL_BUFSZ);
 
         UriParser.extractScheme(getContext().getFileSystemManager().getSchemes(), uri, buf);
-
-        final String[] schemes = getSchemes();
-        for (final String scheme : schemes) {
-            buf.insert(0, ":");
-            buf.insert(0, scheme);
-        }
+        Arrays.stream(getSchemes()).forEach(scheme -> buf.insert(0, scheme+":"));
```
garydgregory, line 58: Formatting: `scheme+":"` -> `scheme + ":"`

PeterAlfredLee: Sure

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/AbstractVfsContainer.java
PeterAlfredLee: It seems the `close` in `java.io.Closeable` may throw an IOException while the `close` in `VfsComponent` won't. This means we have to surround this with a try-catch, which I think is not a good idea.

```
        // Close all components
        Arrays.stream(toclose).filter(component -> component instanceof VfsComponent)
                              .forEach(component -> {
                                  try {
                                      ((java.io.Closeable) component).close();
                                  } catch (IOException e) {
                                      e.printStackTrace();
                                  }
                              });
```

garydgregory: You're right, no need for the extra layer.

PR #97, opened 2020-06-30
[VFS-748] TarProvider Incorrectly marks file IMAGINARY after garbage collection with WeakRefFilesCache

PeterAlfredLee:
Fix for the cache fail in TarFileSystem after garbage collection with weakRefFilesCache. Added a private field cache and removed the local variable strongRef. This is similiar to the cache design in ZipFileSystem.

And also removed the local variable strongRef in ZipFileSystem cause it seems not working at all.


garydgregory:
Hi @PeterAlfredLee 

I could not apply this patch cleanly from a patch file for some reason, but I did bring in the changes with expanded tests. Please see git master and close this PR and https://issues.apache.org/jira/browse/VFS-748 if all is well for your use case.

Thank you!

PeterAlfredLee:
Looks good. Will close the PR and issue.

PR #107, opened 2020-07-30
Fix some test error when jvm's default language is not en

PeterAlfredLee:
Some test's assert expect language is english, and these test are failing when jvm's default language is not en.
This is a fix by setting jvm's default language to en.


garydgregory:
@PeterAlfredLee 
Which asserts do this which which locale? I am wondering if there is an simple fix in the test code. As well, do we need to configure some tests with different different locales instead of forcing single locale build-wide?

PeterAlfredLee:
> Which asserts do this which which locale? 

It's in the `ParseXmlInZipTestCase.testResolveAndParseInvalidXml` and `ParseXmlInZipTestCase.testResolveAndParseNotWellFormedXml`.

> I am wondering if there is an simple fix in the test code. 

I think it's possible. Will push later.

PR #108, opened 2020-07-30
Add proxy config for some Http/Https test

PeterAlfredLee:
Some tests are failing because of the absence of the Http/Https proxy configure.
This is a fix for them.

### DIFF
--- a/commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/http/test/GetContentInfoFunctionalTest.java
+++ b/commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/http/test/GetContentInfoFunctionalTest.java
@@ -20,9 +20,14 @@ import org.apache.commons.vfs2.FileContent;
 import org.apache.commons.vfs2.FileObject;
 import org.apache.commons.vfs2.FileSystemException;
 import org.apache.commons.vfs2.FileSystemManager;
+import org.apache.commons.vfs2.FileSystemOptions;
 import org.apache.commons.vfs2.VFS;
+import org.apache.commons.vfs2.provider.http.HttpFileSystemConfigBuilder;
 import org.junit.Assert;
 import org.junit.Test;
+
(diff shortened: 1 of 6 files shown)


garydgregory:
Hi @PeterAlfredLee 

What does this fix? The Java 8 & 11 builds are green on Travis: https://travis-ci.org/github/apache/commons-vfs/builds/713326898

Java 14 and 15 are failing due to WebDav and JackRabbit 1.x but that is not what you fixing here.

PeterAlfredLee:
Hi @garydgregory 
This is a PR for those machines behind a proxy - some tests are failing because of the lack of the proxy configure for these tests.

ecki:
I think the change is fine, imwould maybe centralize the config helper function, but there is quite a bit of redundancy in the http providers,anyways

Alternative fix would be to stark a mock test server locally.

garydgregory:
@PeterAlfredLee 
Yes, please refactor the duplicate code, I think we can merge this in after.

PeterAlfredLee:
Hi, @garydgregory @ecki 

I agree with you two about the duplicate code. But it seems the `setProxyHost` and `setProxyPort` in `HttpFileSystemConfigBuilder`, `Http4FileSystemConfigBuilder` and `Http5FileSystemConfigBuilder` are not inherited from their common father class or any other interface class. In other words, they implemented the `setProxyHost` and `setProxyPort` separately but only share the same method name.

This make it a bit difficult for me to extract the duplicate code, I can only think of the reflect way to do this([here](https://github.com/PeterAlfredLee/commons-vfs/commit/36e1275c348cc995e55ab48f95e09bfc8bcb6e66)). Do you think this is a good idea? Or you have some better ideas?

garydgregory:
Ah, I see. It's too bad about the shared names outside of an interface or a hierarchy. Reflection is a no-go IMO. I'll review again tomorrow as I think the PR should be OK.

PR #110, opened 2020-08-11
VFS-780 SftpFileSystem returns null channel and produce NPE - fix get…

Wuchte:
Create SftpChannel when it is required

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
@@ -133,7 +133,9 @@ public class SftpFileSystem extends AbstractFileSystem {
                         idleChannel = null;
                     }
                 }
-            } else {
+            }
+
+            if (channel == null) {
                 channel = (ChannelSftp) getSession().openChannel("sftp");
                 channel.connect(connectTimeoutMillis);
                 final Boolean userDirIsRoot = SftpFileSystemConfigBuilder.getInstance()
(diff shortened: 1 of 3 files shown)


garydgregory:
@Wuchte 
Thank you for your PR... BUT... I can't tell what's changed since the whole file is marked here as a diff. May you try to provide a PR that properly shows you diffs please? Also, there is no way to see if the PR fixes anything without a failing unit test. Please see tests we already have for this provider for examples.
Thank you.

Wuchte:
@garydgregory 
Can you take a look patch in [Jira](https://issues.apache.org/jira/browse/VFS-780)?
Something is wrong with github, locally I see correct changeset.

About tests - I'm new in this project, does all providers support parallel access?

garydgregory:
> @garydgregory
> Can you take a look patch in [Jira](https://issues.apache.org/jira/browse/VFS-780)?
> Something is wrong with github, locally I see correct changeset.

Hi @Wuchte 

There is nothing wrong with GitHub; you'll need to consider your line ending settings here: https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings

> 
> About tests - I'm new in this project, does all providers support parallel access?

There is no general rule on multi-threaded access, you'll have to look at the code for each provider and see if they even document their threading model.

Wuchte:
Hi @garydgregory
Can you take a look on fix? EOL is fixed

garydgregory:
Hi @Wuchte 
We need a unit test to show what this fixes. Otherwise, it's just change that happens to not break existing code but there is no proof that it fixes anything.

Wuchte:
some comments about test:
I tried add this test into ProviderWriteTests, but build was failed with timeout(with local debug looks like this scenario doesn't work with "file///..." prefix).
Due this I added this test only for SftpProvider.

Wuchte:
Hi @garydgregory, can you take a look on changes?

garydgregory:
@Wuchte I will take a look tonight.

PR #120, opened 2020-08-30
[VFS-785] Credentials conflicting when root and proxy hosts are same name but different ports

satish-csi:
VFS-785 e.g. if URI for root is https://github.com:8443/test.txt and proxy is github.com:3128 then credentials may get overridden in credential provider map


michael-o (approved):

garydgregory:
@satish-csi Needs unit tests.

PR #137, opened 2020-10-12
Use auto boxing and unboxing

PeterAlfredLee:
We can use auto boxing and auto unboxing to simplify the code.


jochenw:
"We can" doesn't mean "We should".
My personal opinion is, that object creation (Auto boxing) should be clearly visible, because it might have a performance effect.

garydgregory:
> "We can" doesn't mean "We should".
> My personal opinion is, that object creation (Auto boxing) should be clearly visible, because it might have a performance effect.

I agree and I've been saying as much in other PRs within Commons.

garydgregory:
Closing.

PR #139, opened 2020-10-12
Fix test fail caused by lack proxy configuration for machines behind a proxy

PeterAlfredLee:
`org.apache.commons.vfs2.provider.http5s.test.Http5sGetContentInfoTest#testSSLGetContentInfo`  test fail caused by lack proxy configure. This PR is a fix for it.


garydgregory:
Under which circumstance does this test fail? It would be nice to have a better description since the current build is green.

PeterAlfredLee:
Hi @garydgregory 
This is a PR for those machines behind a proxy , same reason as #108

PR #142, opened 2020-10-19
Modify some code use for-each loop and stream API

PeterAlfredLee:
Simplify some code use for-each loop and stream api


garydgregory:
I just started a VOTE thread for 2.7.0-RC1, so I'll take a look at this next week once the release is out.

PR #145, opened 2020-11-04
Modify some calls of method Collection.toArray

PeterAlfredLee:
Use method with empty array instead of pre size array.
Because it's more safe and faster.

See also this [blog](https://shipilev.net/blog/2016/arrays-wisdom-ancients/)


garydgregory:
Now that the blog post is 4 years old, I wonder how current VMs behave... the best thing, for me, that this change has going for it is the simplicity of the pattern; but if you go that route, you should create constants for the empty arrays.

PR #149, opened 2020-12-03
Add ability to remove a provider from DefaultFileSystemManager

boris-petrov:


commons-vfs2/src/main/java/org/apache/commons/vfs2/impl/DefaultFileSystemManager.java
```diff
@@ -208,6 +208,19 @@ public void addProvider(final String[] urlSchemes, final FileProvider provider)
         }
     }
 
+    /**
+     * Unregisters a file system provider.
+     *
+     * @param urlScheme The scheme of the provider.
```
garydgregory, line 214: Missing a since Javadoc tag and a unit test.

commons-vfs2/src/main/java/org/apache/commons/vfs2/impl/DefaultFileSystemManager.java
```diff
@@ -208,6 +208,20 @@ public void addProvider(final String[] urlSchemes, final FileProvider provider)
...
+     * @since 2.8.0
+     */
+    public void removeProvider(final String urlScheme) {
+        final FileProvider provider = providers.remove(urlScheme);
+        // check whether the same instance is not used somewhere else
+        if (provider != null && providers.values().stream().noneMatch(p -> p == provider)) {
+            closeComponent(provider);
```
garydgregory, line 221: Hi @boris-petrov 
Thank you for adding a unit test.
Are you saying here that the same provider can be registered in the providers map more than once but under different keys?
If so, you need to test that case as well.

commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/DefaultFileSystemManagerTest.java
```diff
@@ -47,4 +50,34 @@ public void test_close() throws FileSystemException {
...
     }
 
+    @Test
+    public void testAddAndRemoveProvider() throws FileSystemException {
+        try (DefaultFileSystemManager fileSystemManager = new DefaultFileSystemManager()) {
+            fileSystemManager.setFilesCache(new NullFilesCache());
+            fileSystemManager.setCacheStrategy(CacheStrategy.MANUAL);
+
+            // this is to check when exactly `close` of the provider is called
+            Boolean[] closed = { Boolean.FALSE };
```
garydgregory, line 60: Hi @boris-petrov 
Thank you for your update.
This is a bit of an anti-pattern here, as in general you should use an `AtomicBoolean` instead of a 1-element array. 
BUT, since this is a test and you want to verify that a method was called, this is done using Mockito which we already depend on and you can simply use by following the pattern described here https://stackoverflow.com/questions/9841623/mockito-how-to-verify-method-was-called-on-an-object-created-within-a-method

boris-petrov: I changed the code to use `AtomicBoolean`. However, I'm unsure which pattern exactly you want me to use.

boris-petrov: Actually I did it with Mockito. Check it out.

garydgregory:
@boris-petrov Thank you, merged.

PR #150, opened 2020-12-15
Call `refresh` by default in `AbstractFileObject.onChange`

boris-petrov:
I think this is a good default. Most providers won't do anything different than this. I guess? @garydgregory thoughts?

PR #151, opened 2020-12-16
Refactor getOutputStream

boris-petrov:
I'm not sure whether changing the visibility of `AbstractFileObject.getOutputStream` and marking it as `final` is not a breaking change. This change is not *strictly* needed but I did it to protect from someone accidentally calling it from the outside like was done in `RamFileSystem`. `AbstractFileObject.getOutputStream` is an "internal" method which users should not call because closing the stream returned from it doesn't call the needed hooks - like `endOutput`. That was why `RamFileOutputStream.close` did that explicitly (which is not nice because when creating a new file in the RAM provider, `endOutput` would be called twice).

See also #150.


garydgregory:
Hi @boris-petrov 
Please rebase on git master to pick up the default Maven build for GitHub

boris-petrov:
@garydgregory - done.

garydgregory:
Hi @boris-petrov 
As you can now see in the build output, this PR breaks binary compatibility. We can't do that within a major version, in this case, within the 2.x.y line of releases.
So if you can please fix that, we can then look at the PR afresh.

boris-petrov:
@garydgregory - I rebased and added TODOs instead of changing the methods so that we remember to do it for the next major version.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/AbstractFileObject.java
```diff
@@ -1232,6 +1233,7 @@ public FileName getName() {
         return fileName;
     }
 
+    // TODO: remove this method for the next major version as it is unused
```
garydgregory, line 1236: It's a public method, you do not know who uses it, so you can't say it's "unused". I think this comment can be dropped.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ram/RamFileOutputStream.java
```diff
@@ -92,13 +90,7 @@ public void close() throws IOException {
...
             throw exception;
         }
-        try {
-            this.closed = true;
-            // Close the
-            this.file.endOutput();
-        } catch (final Exception e) {
-            throw new FileSystemException(e);
-        }
+        this.closed = true;
```
garydgregory, line 93: This is a change in behavior that is likely to break some apps since `endOutput()` is no longer called. Needs more study...

garydgregory (changes requested):

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ram/RamFileOutputStream.java
boris-petrov: `endOutput` *will* be called because the only way to get a `RamFileOutputStream` is through `DefaultFileContent` which wraps the output stream and on close calls `endOutput`.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/AbstractFileObject.java
boris-petrov: It's a public method, right. Note, however, that users of VFS always use `FileObject` which is an interface and it doesn't have this method. It's only available if one casts explicitly to `AbstractFileObject` (which no one should do) and to sub-classes (which should not call it at all). The only place this will be called is in `DefaultFileContent` (actually not this but the other overload). That's why I've written that this can be removed - so that in the future a conversation like this is not done. :smiley:

garydgregory: Hi @boris-petrov 
It does not matter as `AbstractFileObject` is a public method.

>if one casts explicitly to AbstractFileObject (which no one should do) 

Again, i does not matter as the class is available for 3rd party providers to subclass. Remember that anyone can plugin their own provider(s) in the framework, it's designed to be extensible. There are two kinds of clients of the API: (1) Traditional consumers of the API, and (2) File System providers, like VFS itself.

For 3.0 we can consider breaking the APIs but we are not there yet.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ram/RamFileOutputStream.java
garydgregory: Hi @rgoers,
Any thoughts as to whether or not this can break existing users?
TY!

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/AbstractFileObject.java
boris-petrov: You are correct for the two usages. My point is that *neither* customers of the API, *nor* FS providers should be calling these methods. Hence they could be made `final` and package-private. But not before 3.0 as you said. And because we're not there, I've added the TODOs so this is not forgotten.

PR #154, opened 2021-01-07
Rework SoftRefFilesCache locking

MaxKellermann:
This PR contains several commits which fix bugs in the SoftRefFilesCache locking code. There are quire a few race conditions and locking bugs. Additionally, it refactors the thread synchronization code to be easier, more robust and faster.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
@@ -237,10 +237,10 @@ public class SoftRefFilesCache extends AbstractFilesCache {
 
     @Override
     public void close() {
-        endThread();
-
         lock.lock();
         try {
+            endThread();
+
             fileSystemCache.clear();
 
             refReverseMap.clear();
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
@@ -96,21 +96,27 @@ public class SoftRefFilesCache extends AbstractFilesCache {
             return;
         }
 
-        synchronized (lock) {
+        lock.lock();
+        try {
             if (softRefReleaseThread == null) {
                 softRefReleaseThread = new SoftRefReleaseThread();
                 softRefReleaseThread.start();
             }
+        } finally {
+            lock.unlock();
         }
     }
 
     private void endThread() {
-        synchronized (lock) {
+        lock.lock();
+        try {
             final SoftRefReleaseThread thread = softRefReleaseThread;
             softRefReleaseThread = null;
             if (thread != null) {
                 thread.interrupt();
             }
+        } finally {
+            lock.unlock();
         }
     }
 
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
@@ -263,7 +263,13 @@ public class SoftRefFilesCache extends AbstractFilesCache {
     @Override
     public void removeFile(final FileSystem fileSystem, final FileName fileName) {
         if (removeFile(new FileSystemAndNameKey(fileSystem, fileName))) {
-            close(fileSystem);
+            lock.lock();
+
+            try {
+                close(fileSystem);
+            } finally {
+                lock.unlock();
+            }
         }
     }
 
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
@@ -43,8 +43,6 @@ import org.apache.commons.vfs2.util.Messages;
  */
 public class SoftRefFilesCache extends AbstractFilesCache {
 
-    private static final int TIMEOUT = 1000;
-
     private static final Log log = LogFactory.getLog(SoftRefFilesCache.class);
 
     private final ConcurrentMap<FileSystem, Map<FileName, Reference<FileObject>>> fileSystemCache = new ConcurrentHashMap<>();
@@ -71,7 +69,7 @@ public class SoftRefFilesCache extends AbstractFilesCache {
         public void run() {
             try {
                 while (!requestEnd) {
-                    final Reference<?> ref = refQueue.remove(TIMEOUT);
+                    final Reference<?> ref = refQueue.remove(0);
                     if (ref == null) {
                         continue;
                     }
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
@@ -88,28 +88,24 @@ public class SoftRefFilesCache extends AbstractFilesCache {
     public SoftRefFilesCache() {
     }
 
+    /**
+     * Called while the lock is held
+     */
     private void startThread() {
-        lock.lock();
-        try {
-            if (softRefReleaseThread == null) {
-                softRefReleaseThread = new SoftRefReleaseThread();
-                softRefReleaseThread.start();
-            }
-        } finally {
-            lock.unlock();
+        if (softRefReleaseThread == null) {
+            softRefReleaseThread = new SoftRefReleaseThread();
+            softRefReleaseThread.start();
         }
     }
 
+    /**
+     * Called while the lock is held
+     */
     private void endThread() {
-        lock.lock();
-        try {
-            final SoftRefReleaseThread thread = softRefReleaseThread;
-            softRefReleaseThread = null;
-            if (thread != null) {
-                thread.interrupt();
-            }
-        } finally {
-            lock.unlock();
+        final SoftRefReleaseThread thread = softRefReleaseThread;
+        softRefReleaseThread = null;
+        if (thread != null) {
+            thread.interrupt();
         }
     }
 
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
@@ -262,17 +262,19 @@ public class SoftRefFilesCache extends AbstractFilesCache {
 
     @Override
     public void removeFile(final FileSystem fileSystem, final FileName fileName) {
-        if (removeFile(new FileSystemAndNameKey(fileSystem, fileName))) {
-            lock.lock();
-
-            try {
+        lock.lock();
+        try {
+            if (removeFile(new FileSystemAndNameKey(fileSystem, fileName))) {
                 close(fileSystem);
-            } finally {
-                lock.unlock();
             }
+        } finally {
+            lock.unlock();
         }
     }
 
+    /**
+     * Called while the lock is held
+     */
     private boolean removeFile(final FileSystemAndNameKey key) {
         if (log.isDebugEnabled()) {
             log.debug("removeFile: " + this.getSafeName(key.getFileName()));
@@ -280,17 +282,12 @@ public class SoftRefFilesCache extends AbstractFilesCache {
 
         final Map<?, ?> files = getOrCreateFilesystemCache(key.getFileSystem());
 
-        lock.lock();
-        try {
-            final Object ref = files.remove(key.getFileName());
-            if (ref != null) {
-                refReverseMap.remove(ref);
-            }
-
-            return files.isEmpty();
-        } finally {
-            lock.unlock();
+        final Object ref = files.remove(key.getFileName());
+        if (ref != null) {
+            refReverseMap.remove(ref);
         }
+
+        return files.isEmpty();
     }
 
     protected Map<FileName, Reference<FileObject>> getOrCreateFilesystemCache(final FileSystem fileSystem) {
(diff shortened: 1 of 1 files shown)


commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
```diff
@@ -43,85 +38,55 @@
...
 
-    private final ConcurrentMap<FileSystem, Map<FileName, Reference<FileObject>>> fileSystemCache = new ConcurrentHashMap<>();
+    private final Map<FileSystem, Map<FileName, Reference<FileObject>>> fileSystemCache = new HashMap<>();
     private final Map<Reference<FileObject>, FileSystemAndNameKey> refReverseMap = new HashMap<>(100);
     private final ReferenceQueue<FileObject> refQueue = new ReferenceQueue<>();
 
-    private volatile SoftRefReleaseThread softRefReleaseThread = null; // @GuardedBy("lock")
-
-    private final Lock lock = new ReentrantLock();
+    private SoftRefReleaseThread softRefReleaseThread = null;
```
garydgregory, line 47: Please don't initialize to default values.

garydgregory (changes requested):
Hi Max,
Thank you for you PR.

You mention that there are bugs, ok. But what are they? At the very least we need a list so users can know if a new release addresses an issue they may have. 

Without unit tests that fail without the main changes, the code may just be changed and regress to the old behavior in the future, only to be fixed again later, round and round.

I call these "blind" PRs because we can't even tell if the bug or bugs is actually fixed. 

I am quite aware that creating tests for a synchronization issue can be challenging, so I understand the desire to just fix it in the main code and move on. 

May you please:
- document here what bugs you are fixing
- document in the code why the code is the way it is
- create unit tests, otherwise don't be surprised if code related to PRs like these are either changed, break, or revert to other unwanted behavior, or not even merged, which is counterproductive: -(

TY!

commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
MaxKellermann: Agree, but not my code / choice. I only removed the `volatile` and the obsolete code comment. I preserved the old code as-is, and I think refactoring to remove unnecessary default values should not be mixed in this commit.

MaxKellermann:
> You mention that there are bugs, ok. But what are they? At the very least we need a list so users can know if a new release addresses an issue they may have.

Each commit which addresses a bug describes it in the commit message.

In particular, those 4 commits address real bugs, each documented:
https://github.com/apache/commons-vfs/pull/154/commits/98d646e95eed47b11bdb063911161b667223df51
https://github.com/apache/commons-vfs/pull/154/commits/f210ac6f2a338f2aeb18217b3c408c1dd92eacca
https://github.com/apache/commons-vfs/pull/154/commits/eae0cd551978a83ffaeffa77622a8dbc8762acb0
https://github.com/apache/commons-vfs/pull/154/commits/b946d8077a686fc2af236b81a56f55db74e727c0

The rest is just refactoring and optimization.

> Without unit tests that fail without the main changes, the code may just be changed and regress to the old behavior in the future, only to be fixed again later, round and round.

How do you unit-test race conditions / TOCTOU bugs?

I had a look at the current unit tests for this class - and despite it being allegedly thread-safe, there is not a single unit test testing anything about thread-safety. Hey, the unit tests are so short that I can paste *all of it* here:
https://github.com/apache/commons-vfs/blob/3916dd2a2f88c9ae27d040acc11c63eeb7b1ca64/commons-vfs2/src/test/java/org/apache/commons/vfs2/cache/SoftRefFilesCacheTests.java#L31-L35
That's all!

And this is a class whose `SoftRefReleaseThread` was broken for 5 years (due to an inverted check added by the completely-bad-commit 9e360ca017b1f639957c799a8b7b64047c286d93) that the code could *never* run, leading to an unbounded memory leak. This was fixed by commit https://github.com/apache/commons-vfs/commit/f09035290b1cdcf54f40c331e8c84d925eb454b7 which introduced a new bug (fixed by this PR), but without any unit tests (and with a very confusing commit message which missed the point).

If there was an example in this library's unit tests on how to write a unit test for such a bug, I could give it a try, but I see none. This would be the first time such a unit test for a race condition fix is required.

But yes, basically we agree that having a unit test would be great, but my motivation to figure this out from scratch doesn't exist.

MaxKellermann:
Rebased and added an extra commit removing the unnecessary initializer, as requested. (This initializer has been there before, I did not add it, and it has actually nothing to do with this PR.)

MaxKellermann:
Rebased and fixed conflict because the meanwhile-merged commit 048d34d1e0de11c1ecf44efcae58a225cee1ea96 duplicates my commit "SoftRefFilesCache: remove unnecessary initializer" which I authored after it was requested by @garydgregory ... which has now turned out to be a waste of time, because somebody else's version authored after mine was merged instead.

garydgregory:
@ecki , @rgoers, any thoughts?

commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
```diff
@@ -290,6 +286,9 @@ private boolean removeFile(final FileSystemAndNameKey key) {
         return files.isEmpty();
     }
 
+    /**
+     * Called while the lock is held
+     */
     protected Map<FileName, Reference<FileObject>> getOrCreateFilesystemCache(final FileSystem fileSystem) {
```
efge: Could you make this `private` so that it's consistent with all the other methods that are "called while the lock is held"?

efge:
Given that the Loom project will likely have issues with `synchronized` in its first iterations, I would not include the last commit switching from locks to synchronized. Otherwise the cleanups look good.

I totally sympathise with @garydgregory about the lack of test, but also with @MaxKellermann in that writing tests that exercise thread-safety issues is a PITA.

MaxKellermann:
> Given that the Loom project will likely have issues with `synchronized` in its first iterations

What kind of problems are caused by `synchronized`? I'm not a Java expert, I'm curious, I thought that was the first-class synchronization feature that always "just works", and you only need stuff like `ReentrantLock` when you need special features.

kinow:
>I totally sympathise with @garydgregory about the lack of test, but also with @MaxKellermann in that writing tests that exercise thread-safety issues is a PITA.

Ditto here. Writing tests for the thread-safety would be nice, but understandable if it's too hard to write.

@MaxKellermann, @garydgregory what do you think if we break this PR into a few other smaller PR's, I think that would be easier to review, and grouping that way in case some bugs/issue is raised later, it might be easier to look at the changelog and see what could be causing it.

For example, I think we could have one PR with smal improvements like (easy to review, don't think we need tests):

- remove `super.close()` (maybe add a comment to the code where it was removed so we don't accidentally add it back? Will leave a comment there...)
- use isEmpty() instead of size < 1
- remove redundant isInterrupted() check 
- simplify the InterruptedException catch block 

Another for the Lock issues (great catch! in some cases we locked twice, in others forgot to require the lock, and in other cases we also locked the Lock object without acquiring the lock; I'd say we could code-review this change without a test [testing synchronized code blocks is hard, but code that is acquiring a Lock is even harder IMO])

- fix ReentrantLock usage in {start,end}Thread() 
- add missing lock to removeFile() 
- require lock while calling removeFile(FileSystemAn… 
- require lock while calling getOrCreateFilesystemCa… 
- move endThread() call inside the lock 
- require lock for startThread(), endThread() 
- move code to removeFile(Reference)
- don't use ConcurrentMap (oh, I liked this! Would need to read the code with calm to understand if there was no other case for the concurrent map, but I think you might be correct here!)

Another PR for the volatile change (code-review might do I think)

- eliminate "volatile" on softRefReleaseThread 

And one more for the removal of `ReentrantLock`? Or maybe just create an issue for discussion regarding the Loom project with fibers and possible issues (I think that's what @efge meant?).

kinow (commented):
>I totally sympathise with @garydgregory about the lack of test, but also with @MaxKellermann in that writing tests that exercise thread-safety issues is a PITA.

Ditto here. Writing tests for the thread-safety would be nice, but understandable if it's too hard to write.

@MaxKellermann, @garydgregory what do you think if we break this PR into a few other smaller PR's, I think that would be easier to review, and grouping that way in case some bugs/issue is raised later, it might be easier to look at the changelog and see what could be causing it.

For example, I think we could have one PR with smal improvements like (easy to review, don't think we need tests):

- remove `super.close()` (maybe add a comment to the code where it was removed so we don't accidentally add it back? Will leave a comment there...)
- use isEmpty() instead of size < 1
- remove redundant isInterrupted() check 
- simplify the InterruptedException catch block 

Another for the Lock issues (great catch! in some cases we locked twice, in others forgot to require the lock, and in other cases we also locked the Lock object without acquiring the lock; I'd say we could code-review this change without a test [testing synchronized code blocks is hard, but code that is acquiring a Lock is even harder IMO])

- fix ReentrantLock usage in {start,end}Thread() 
- add missing lock to removeFile() 
- require lock while calling removeFile(FileSystemAn… 
- require lock while calling getOrCreateFilesystemCa… 
- move endThread() call inside the lock 
- require lock for startThread(), endThread() 
- move code to removeFile(Reference)
- don't use ConcurrentMap (oh, I liked this! Would need to read the code with calm to understand if there was no other case for the concurrent map, but I think you might be correct here!)

Another PR for the volatile change (code-review might do I think)

- eliminate "volatile" on softRefReleaseThread 

And one more for the removal of `ReentrantLock`? Or maybe just create an issue for discussion regarding the Loom project with fibers and possible issues (I think that's what @efge meant?).

MaxKellermann:
> @MaxKellermann, @garydgregory what do you think if we break this PR into a few other smaller PR's, I think that would be easier to review, and grouping that way in case some bugs/issue is raised later, it might be easier to look at the changelog and see what could be causing it.

If that makes it easier for you, OK for me. Here's part 1: https://github.com/apache/commons-vfs/pull/158
As soon as you merge it, I'll rebase this PR and drop the merged commits.

In projects I maintain, if I like some obvious-good commits, I cherry-pick them and then start discussing the others, and I expect the submitter to rebase the PR with only the unmerged commits remaining.

What I certainly would never do is fold all PR commits into one large commit like you did with https://github.com/apache/commons-vfs/pull/155 - 3 commits were folded into one (https://github.com/apache/commons-vfs/commit/9c4302e4559acb1e2a627f57f49105c54fe09895). That loses a lot of context, makes the resulting commit hard to read/verify/understand, and makes git-bisect very hard. That's your project, your maintenance decision, so I'm not going to argue, just mentioning it because that's one thing that would make project management hard for me.

kinow:
>If that makes it easier for you, OK for me. Here's part 1: #158
>As soon as you merge it, I'll rebase this PR and drop the merged commits.

Thanks @MaxKellermann !

>In projects I maintain, if I like some obvious-good commits, I cherry-pick them and then start discussing the others, and I expect the submitter to rebase the PR with only the unmerged commits remaining.

:+1: 

>What I certainly would never do is fold all PR commits into one large commit like you did with #155 - 3 commits were folded into one (9c4302e). That loses a lot of context, makes the resulting commit hard to read/verify/understand, and makes git-bisect very hard. That's your project, your maintenance decision, so I'm not going to argue, just mentioning it because that's one thing that would make project management hard for me.

My workflow is normally to check out the branch locally, rebase onto master — or make sure it's rebased — and then checkout master and merge with `--no-ff`. If the PR has many commits, I may squash them if they appear to be draft/WIP commits, or ask the user if s/he can squash it (asked last week I think, in Apache Commons Imaging).

But sometimes we may squash without thinking/asking. Some times when I create a PR I mention that I haven't squashed for some specific reason, and tell the maintainers I'd be happy to squash it if they think it's necessary. If you mention that I'm sure the final reviewer will consider it too.

Some projects also have a hard-policy to always merge a single commit (Apache OpenNLP for example, though not sure if that changed).

Finally, I forgot to say **thank you** for the great commit messages. They are short, and explain what's included in that commit :+1:

efge:
> > Given that the Loom project will likely have issues with `synchronized` in its first iterations
> 
> What kind of problems are caused by `synchronized`? I'm not a Java expert, I'm curious, I thought that was the first-class synchronization feature that always "just works", and you only need stuff like `ReentrantLock` when you need special features.

It is today, but the ongoing Loom project aims at adding new lightweight thread scheduling features for which the low-level monitor that `synchronized` takes is harder to interact with (again: in the first iterations, it'll eventually work ok). See https://wiki.openjdk.java.net/display/loom/Main for more, search for "synchronized".

Therefore the Java libraries that aim to be Loom-friendly try to currently avoid `synchronized` blocks. Here's an example from the PostgreSQL JDBC driver: https://github.com/pgjdbc/pgjdbc/issues/1951

ecki:
Personally I don’t care much for this rework, synchronized blocks are much easier to read and Loom will have to find a solution for it, anyway. There are better reasons, like having finer control and try and statistics, but I don’t think this applies here. What I will think is much more important is to,actually define the concurrency model and document it

MaxKellermann:
> Loom will have to find a solution for it, anyway.

I agree with that. I read that OpenJDK link, and my first thought was: this is a bad implementation if it makes `synchronized` so special that it blocks context switches. There's no reason why it would need to, that's just an (unoptimized) implementation detail.

If this Loom thing forces everybody to rewrite their code, replacing a simple thing as `synchronized` with more complicated, error-prone code, it's bad. But I can't decide what you want to support, it's not my project here ...

garydgregory:
I do not think we should twist ourselves into a pretzel over project Loom now. It's way too soon.

MaxKellermann:
Rebased after #158 was merged, dropping its 4 commits.

garydgregory:
> > > Given that the Loom project will likely have issues with `synchronized` in its first iterations
> > 
> > 
> > What kind of problems are caused by `synchronized`? I'm not a Java expert, I'm curious, I thought that was the first-class synchronization feature that always "just works", and you only need stuff like `ReentrantLock` when you need special features.
> 
> It is today, but the ongoing Loom project aims at adding new lightweight thread scheduling features for which the low-level monitor that `synchronized` takes is harder to interact with (again: in the first iterations, it'll eventually work ok). See https://wiki.openjdk.java.net/display/loom/Main for more, search for "synchronized".
> 
> Therefore the Java libraries that aim to be Loom-friendly try to currently avoid `synchronized` blocks. Here's an example from the PostgreSQL JDBC driver: [pgjdbc/pgjdbc#1951](https://github.com/pgjdbc/pgjdbc/issues/1951)

I do not think that "Loom-friendliness" is a goal in this component; we are, after all, still on Java 8. Java 11 is not even running "everywhere" yet, for example, on i/Series.

garydgregory:
Hi All,

I plan on cutting a release candidate for VFS this week (maybe even tonight). If someone wants to review and merge this or not, go for it. Otherwise, it can go in the next release, RERO :-)

garydgregory:
I want to move this discussion to the dev mailing list. Please post there.

PR #155, opened 2021-01-13
FileContentThreadData: lazy ArrayList initialization to save memory

MaxKellermann:
Don't create ArrayList instances that are never used.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/DefaultFileContent.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/DefaultFileContent.java
@@ -442,7 +442,7 @@ public final class DefaultFileContent implements FileContent {
             final FileContentThreadData threadData = getFileContentThreadData();
 
             // Close the input stream
-            while (threadData.getInstrsSize() > 0) {
+            while (threadData.hasInputStream()) {
                 final InputStream inputStream = threadData.removeInputStream(0);
                 try {
                     if (inputStream instanceof FileContentInputStream) {
@@ -459,7 +459,7 @@ public final class DefaultFileContent implements FileContent {
             }
 
             // Close the randomAccess stream
-            while (threadData.getRastrsSize() > 0) {
+            while (threadData.hasRandomAccessContent()) {
                 final FileRandomAccessContent randomAccessContent = (FileRandomAccessContent) threadData
                         .removeRastr(0);
                 try {
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/FileContentThreadData.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/FileContentThreadData.java
@@ -58,8 +58,8 @@ class FileContentThreadData {
         this.randomAccessContentList.add(randomAccessContent);
     }
 
-    int getInstrsSize() {
-        return this.inputStreamList.size();
+    boolean hasInputStream() {
+        return !this.inputStreamList.isEmpty();
     }
 
     public Object removeInstr(final int pos) {
@@ -91,7 +91,7 @@ class FileContentThreadData {
         outputStream = null;
     }
 
-    int getRastrsSize() {
-        return randomAccessContentList.size();
+    boolean hasRandomAccessContent() {
+        return !randomAccessContentList.isEmpty();
     }
 }
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/FileContentThreadData.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/FileContentThreadData.java
@@ -83,7 +83,7 @@ class FileContentThreadData {
     }
 
     public boolean hasStreams() {
-        return !inputStreamList.isEmpty() || outputStream != null || !randomAccessContentList.isEmpty();
+        return hasInputStream() || outputStream != null || hasRandomAccessContent();
     }
 
     public void closeOutstr() throws FileSystemException {
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/FileContentThreadData.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/FileContentThreadData.java
@@ -29,8 +29,8 @@ class FileContentThreadData {
 
     // private int state = DefaultFileContent.STATE_CLOSED;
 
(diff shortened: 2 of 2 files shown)


MaxKellermann:
Rebased (and fixed conflicts).

boris-petrov:
@MaxKellermann, @garydgregory - this commit (I think) causes my tests to fail with NPEs on version 2.8.0:

```
Caused by: java.lang.NullPointerException: Cannot invoke "java.util.ArrayList.remove(Object)" because "this.inputStreamList" is null
	at org.apache.commons.vfs2.provider.FileContentThreadData.remove(FileContentThreadData.java:82)
	at org.apache.commons.vfs2.provider.DefaultFileContent.endInput(DefaultFileContent.java:555)
	at org.apache.commons.vfs2.provider.DefaultFileContent.access$000(DefaultFileContent.java:44)
	at org.apache.commons.vfs2.provider.DefaultFileContent$FileContentInputStream.onClose(DefaultFileContent.java:662)
	at org.apache.commons.vfs2.util.MonitorInputStream.close(MonitorInputStream.java:92)
	at org.apache.commons.vfs2.provider.DefaultFileContent$FileContentInputStream.close(DefaultFileContent.java:648)
```

Isn't there a test that gets a file's content and then closes it right after that? I think that will reproduce the problem.

MaxKellermann:
@boris-petrov I did not add a `null` check to `remove()` because it should be impossible to call `remove()` without previously having called `add()` (which initializes the field).
But if it is indeed possible (which your crash shows), then this means that `remove()` has been buggy all along: previously, the bug was just hidden because `this.inputStreamList.remove(inputStream);` did not have an effect - the given object was not there, because it was in another thread's `FileContentThreadData` instance.

boris-petrov:
@MaxKellermann - is this something that you'll look into? Can you reproduce the problem - that is, do you have an idea what is going on?

MaxKellermann:
No, I can't reproduce the problem (because I don't use commonfs-vfs).
How do you? Do you invoke "open" and "close" on a stream from different threads?

boris-petrov:
Yes, this is exactly what happens.

MaxKellermann:
OK, it looks like the commons-vfs code assumes that any stream object must be closed by the same thread that opened it. But this requirement doesn't appear to be documented, and doesn't make a lot of sense - why not move a stream object to another thread? A sane library should allow that, but as I said: commons-vfs has been buggy all along, and my patch only revealed that existing bug.
I could easily add a `null` check which would make the symptom go away, but this doesn't solve the real bug, so I don't like doing that. On the other hand, fixing the real problem may mean removing the whole ThreadLocal logic, because it's inherently broken.
@garydgregory what is your opinion on this?

boris-petrov:
I vote with 2 hands and 2 legs for removing the whole ThreadLocal horror. This is bad practice anyway and won't work on Loom nicely. Do you have understanding of the code - what exactly this thread local is used for?

MaxKellermann:
It is used to keep track of whether the current thread has a stream object for a certain file (`DefaultFileContent`) open. This is needed to implement `FileContent.isOpen()`, which in turn gets called by `AbstractFileContent.isContentOpen()`, which is only used by `LRUFilesCache` to check whether eviction shall be blocked. Don't ask me why it only checks for stream objects of the current thread... no idea. None of this makes a lot of sense to me, it seems like a bad hack.

boris-petrov:
So it seems, yes. The good thing is that it sounds straightforward to fix (i.e. remove the thread local and use a volatile boolean or something) or am I missing something?

MaxKellermann:
Being a C++ expert, I have very little experience with Java's memory model, but in C or C++, `volatile` is not thread-safe (even though many people believe it is, and get away with the bad assumption most of the time). Apart from that, tracking users of an object across all threads is one thing - but making a decision based on that in `LRUFilesCache` is even harder, due to potential TOCTOU problems.

boris-petrov:
I'm not really familiar with the code, but looking through it I see a few funny things. `FileContentThreadData` is explicitly [marked in a comment](https://github.com/apache/commons-vfs/blob/rel/commons-vfs-2.8.0/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/FileContentThreadData.java#L26) that the data it holds should be for a single thread only. It [says the same](https://github.com/apache/commons-vfs/blob/rel/commons-vfs-2.8.0/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/DefaultFileContent.java#L598) for `DefaultFileContent.isOpen`. I don't understand what's the point of that or is there even a point. Implementing a global `isOpen` check for all threads should be trivial enough with some collections which keep track of the open streams and some locking here and there.

I can't say if that will be a problem for `LRUFilesCache` but it's definitely going to be better than the current state.

@garydgregory what do you think?

garydgregory:
Well, VFS is a pretty old library, and while there are allowances for multi-threading, I think it is fair to say that the each provider should be treated separately for thread safety assurances and that in general, VFS is should not be considered thread safe unless the docs for a specific class or method says so. 

IOW, thread safety is a feature you do NOT get by default. 

I am open to review patches (already merged one yesterday) and to release a 2.8.1 fairly soon.

Recall that the default build must pass.

MaxKellermann:
@garydgregory this isn't about thread-safety as in "concurrent use"; this (old, now-revealed) bug is about "create a stream in one thread and close it in another (but no concurrent use)".
The bug has always been there and affects *all* providers - more exactly: all providers which use `AbstractFileObject` as base class (which is: all of them, isn't it?).

Previously, the bug led to problems which were mostly invisible; i.e. the `FileContentThreadData` silently leaked objects. Now it's a crash (`NullPointerException`).

As I said, the `NullPointerException` can be made go away easily by adding a `null` check, but the underlying bug will still be there.

However, if you decide that commons-vfs is not thread-safe as in "stream objects must not be passed across thread boundaries", we don't need to do anything. Then the crash @boris-petrov faces is a bug in his own code, a wrong use of the commons-vfs API.

Before anybody can attempt to fix this, an authority for commons-vfs must decide what is an allowed use of the API. If there are restrictions on passing objects across thread boundaries, those should be documented. Only after this documentation exists, we can come up with a solution.

boris-petrov:
@MaxKellermann - I agree with most of the things you say except that an "authority for commons-vfs must decide" on something. This is an open source project and PRs are accepted. If we fix the bug transparently for everybody and allow use-cases like mine - by removing the ThreadLocal and implementing `isOpen` in a different way - then this is best for everyone.

PR #158, opened 2021-02-06
Rework SoftRefFilesCache locking part 1

MaxKellermann:
Splitted out of https://github.com/apache/commons-vfs/pull/154 as suggested by @kinow

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/cache/SoftRefFilesCache.java
@@ -69,7 +69,7 @@ public class SoftRefFilesCache extends AbstractFilesCache {
 
         @Override
         public void run() {
-            loop: while (!requestEnd && !Thread.currentThread().isInterrupted()) {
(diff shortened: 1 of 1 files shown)


kinow (approved):
Looks good to me! Had to open the `SoftRefFilesCache` side by side to look at the named-loop that was removed, but I think it is doing the same thing, just simpler.

I think we only need a JIRA for this, add to the changelog, and then it should be ready to be merged. @MaxKellermann we normally ask contributors to create the issue, but let's wait for others to read the original PR, and see if they agree on breaking it down into smaller PR's.

If so I can create the JIRA issue and add the changelog here after they finish reviewing it :+1: 

Thanks!

MaxKellermann:
@garydgregory?

MaxKellermann:
Thanks @garydgregory, I'll rebase #154 now.

kinow:
Created and linked issue in JIRA https://issues.apache.org/jira/browse/VFS-795

Added changelog in our `changes.xml` https://github.com/apache/commons-vfs/commit/a1a55155e1c8889eb57cc9cf8df58091fad2c4c5

Thanks @MaxKellermann !

PR #161, opened 2021-03-02
Avoid File Stream

arturobernalg:
Use Files.newOutputStream instead of new FileOutputStream (fileName).
Use Files.newInputStream instead of new  FileInputStream  (fileName)


arturobernalg:
I don't know if mi PR broke the test or there are some problem with travis. i appreciate a little help here,

Thanks

garydgregory:
The master branch is building fine here in GitHub Actions, so it must be something wrong in your branch. You'll need to run your code in a debugger and see what happens.

arturobernalg:
> The master branch is building fine here in GitHub Actions, so it must be something wrong in your branch. You'll need to run your code in a debugger and see what happens.

Thank you @garydgregory  I'll try to check what happened later.

PR #164, opened 2021-03-11
Replace construction of FileInputStream and FileOutputStream objects with Files NIO APIs.

arturobernalg:
Use Files.newOutputStream instead of new FileOutputStream (fileName).
Use Files.newInputStream instead of new FileInputStream (fileName)

### DIFF
--- a/commons-vfs2-jackrabbit1/src/test/java/org/apache/commons/vfs2/provider/webdav/test/WebdavProviderTestCase.java
+++ b/commons-vfs2-jackrabbit1/src/test/java/org/apache/commons/vfs2/provider/webdav/test/WebdavProviderTestCase.java
@@ -17,9 +17,9 @@
 package org.apache.commons.vfs2.provider.webdav.test;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.Files;
 
 import javax.jcr.Node;
 import javax.jcr.NodeIterator;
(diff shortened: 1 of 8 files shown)


arturobernalg:
> What about elsewhere? Like the tests. Should those be updated as well?

HI @garydgregory 

I think it's already cover and not need additional test.What do you think?
https://github.com/apache/commons-vfs/blob/master/commons-vfs2/src/test/java/org/apache/commons/vfs2/ProviderWriteAppendTests.java#L71

garydgregory:
> > What about elsewhere? Like the tests. Should those be updated as well?
> 
> HI @garydgregory
> 
> I think it's already cover and not need additional test.What do you think?
> https://github.com/apache/commons-vfs/blob/master/commons-vfs2/src/test/java/org/apache/commons/vfs2/ProviderWriteAppendTests.java#L71

@arturobernalg 
I have not talking about code coverage. I am referring to other call sites, for example `org.apache.commons.vfs2.filter.BaseFilterTest.zipDir(File, FileFilter, String, File)` and others.

arturobernalg:
> > > What about elsewhere? Like the tests. Should those be updated as well?
> > 
> > 
> > HI @garydgregory
> > I think it's already cover and not need additional test.What do you think?
> > https://github.com/apache/commons-vfs/blob/master/commons-vfs2/src/test/java/org/apache/commons/vfs2/ProviderWriteAppendTests.java#L71
> 
> @arturobernalg
> I was not talking about code coverage. I am referring to other call sites, for example `org.apache.commons.vfs2.filter.BaseFilterTest.zipDir(File, FileFilter, String, File)` and others.

Sorry @garydgregory 
I was not understanding you. Changed

TY

PR #166, opened 2021-03-11
Remove usage of `ThreadLocal` in `DefaultFileContent`

boris-petrov:
This also fixes inability to close a stream from a different thread than the one that opened it. See discussion in https://github.com/apache/commons-vfs/pull/155.

I tried to mimic pretty much exactly the code before (to minimize the changes) but to fix the underlying issue. There are still problems like concurrently closing one stream (which potentially does `this.fileContentThreadData = null` if it was the last stream) and opening a new one (which might add a stream to the existing `this.fileContentThreadData`, causing it to be "lost" because of the setting of the variable to `null` during the closing) but 1) they were also there before; 2) they will be very rare and 3) should be the subject of a later PR.

cc @garydgregory, @MaxKellermann

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/FileContentThreadData.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/FileContentThreadData.java
@@ -23,7 +23,7 @@ import org.apache.commons.vfs2.FileSystemException;
 import org.apache.commons.vfs2.RandomAccessContent;
 
 /**
- * Holds the data which needs to be local to the current thread
+ * Holds information about the opened streams of a file (from any thread)
  */
 class FileContentThreadData {
 
--- a/commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/DefaultFileContentTest.java
+++ b/commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/DefaultFileContentTest.java
@@ -16,6 +16,7 @@
  */
 package org.apache.commons.vfs2.provider;
 
+import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -26,6 +27,7 @@ import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.vfs2.FileContent;
 import org.apache.commons.vfs2.FileObject;
+import org.apache.commons.vfs2.FileSystemException;
 import org.apache.commons.vfs2.FileSystemManager;
 import org.apache.commons.vfs2.VFS;
 import org.junit.Assert;
@@ -77,6 +79,11 @@ public class DefaultFileContentTest {
     @Test(expected = IllegalArgumentException.class)
     public void testInputStreamBufferSizeNegative() throws Exception {
         testInputStreamBufferSize(-2);
+    }
+
+    @Test
+    public void testInputStreamClosedInADifferentThread() throws Exception {
+        testStreamClosedInADifferentThread(content -> content.getInputStream());
     }
 
     @Test
@@ -167,4 +174,32 @@ public class DefaultFileContentTest {
         }
     }
 
+    @Test
+    public void testOutputStreamClosedInADifferentThread() throws Exception {
+        testStreamClosedInADifferentThread(content -> content.getOutputStream());
+    }
+
+    private <T extends Closeable> void testStreamClosedInADifferentThread(FunctionThrowingException<FileContent, T> getStream) throws Exception {
+        final File temp = File.createTempFile("temp-file-name", ".tmp");
+        final FileSystemManager fileSystemManager = VFS.getManager();
+
+        try (FileObject file = fileSystemManager.resolveFile(temp.getAbsolutePath())) {
+            T stream = getStream.apply(file.getContent());
+            boolean[] check = { false };
+            Thread thread = new Thread(() -> {
+                try {
+                    stream.close();
+                } catch (IOException exception) {
+                }
+                check[0] = true;
+            });
+            thread.start();
+            thread.join();
+            Assert.assertTrue(check[0]);
+        }
+    }
+}
+
+interface FunctionThrowingException<T, R> {
+    R apply(T t) throws FileSystemException;
 }
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/DefaultFileContent.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/DefaultFileContent.java
@@ -62,8 +62,8 @@ public final class DefaultFileContent implements FileContent {
     private Map<String, Object> roAttrs;
     private FileContentInfo fileContentInfo;
     private final FileContentInfoFactory fileContentInfoFactory;
+    private FileContentThreadData fileContentThreadData;
 
-    private final ThreadLocal<FileContentThreadData> threadLocal = ThreadLocal.withInitial(FileContentThreadData::new);
     private boolean resetAttributes;
 
     /**
@@ -76,8 +76,15 @@ public final class DefaultFileContent implements FileContent {
         this.fileContentInfoFactory = fileContentInfoFactory;
     }
 
+    /**
+     * This method should *only* be called in a synchronized context to prevent race-conditions! See:
+     * https://github.com/apache/commons-vfs/pull/166
+     */
     private FileContentThreadData getFileContentThreadData() {
-        return this.threadLocal.get();
+        if (fileContentThreadData == null) {
+            fileContentThreadData = new FileContentThreadData();
+        }
+        return fileContentThreadData;
     }
 
     void streamOpened() {
@@ -374,7 +381,9 @@ public final class DefaultFileContent implements FileContent {
 
         final FileRandomAccessContent rac = new FileRandomAccessContent(fileObject, rastr);
 
-        getFileContentThreadData().add(rac);
+        synchronized (this) {
+            getFileContentThreadData().add(rac);
+        }
         streamOpened();
 
         return rac;
@@ -438,12 +447,29 @@ public final class DefaultFileContent implements FileContent {
     @Override
     public void close() throws FileSystemException {
         FileSystemException caught = null;
-        try {
-            final FileContentThreadData threadData = getFileContentThreadData();
 
-            // Close the input stream
-            while (threadData.hasInputStream()) {
-                final InputStream inputStream = threadData.removeInputStream(0);
+        while (true) {
+            InputStream inputStream = null;
+            FileRandomAccessContent randomAccessContent = null;
+            FileContentOutputStream outputStream = null;
+
+            synchronized (this) {
+                final FileContentThreadData threadData = getFileContentThreadData();
+
+                if (threadData.hasInputStream()) {
+                    inputStream = threadData.removeInputStream(0);
+                } else if (threadData.hasRandomAccessContent()) {
+                    randomAccessContent = (FileRandomAccessContent) threadData.removeRandomAccessContent(0);
+                } else if (threadData.getOutputStream() != null) {
+                    outputStream = threadData.getOutputStream();
+                    threadData.setOutputStream(null);
+                } else {
+                    fileContentThreadData = null;
+                    break;
+                }
+            }
+
+            if (inputStream != null) {
                 try {
                     if (inputStream instanceof FileContentInputStream) {
                         ((FileContentInputStream) inputStream).close();
@@ -454,33 +480,20 @@ public final class DefaultFileContent implements FileContent {
                     }
                 } catch (final FileSystemException ex) {
                     caught = ex;
-
                 }
-            }
-
-            // Close the randomAccess stream
-            while (threadData.hasRandomAccessContent()) {
-                final FileRandomAccessContent randomAccessContent = (FileRandomAccessContent) threadData
-                        .removeRandomAccessContent(0);
+            } else if (randomAccessContent != null) {
                 try {
                     randomAccessContent.close();
                 } catch (final FileSystemException ex) {
                     caught = ex;
                 }
-            }
-
-            // Close the output stream
-            final FileContentOutputStream outputStream = threadData.getOutputStream();
-            if (outputStream != null) {
-                threadData.setOutputStream(null);
+            } else if (outputStream != null) {
                 try {
                     outputStream.close();
                 } catch (final FileSystemException ex) {
                     caught = ex;
                 }
             }
-        } finally {
-            threadLocal.remove();
         }
 
         // throw last error (out >> rac >> input) after all closes have been tried
@@ -517,7 +530,9 @@ public final class DefaultFileContent implements FileContent {
                     : new FileContentInputStream(fileObject, inputStream, bufferSize);
             // @formatter:on
         }
-        getFileContentThreadData().add(wrappedInputStream);
+        synchronized (this) {
+            getFileContentThreadData().add(wrappedInputStream);
+        }
         streamOpened();
 
         return wrappedInputStream;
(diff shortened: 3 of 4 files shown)


garydgregory:
Hi @boris-petrov 
Thank you for PR.
May you please rebase on `master` to pick up the latest to make sure we are still green.

MaxKellermann:
Previously, at most one `OutputStream` was allowed *per thread*. With your change, at most one `OutputStream` is allowed *globally*. This might break applications.

garydgregory:
Hi @boris-petrov 

> This also fixes inability to close a stream from a different thread

That's one you can write a test to prove the fix actually works.

MaxKellermann:
One more bug: TOCTOU bug in all functions which set `this.fileContentThreadData = null` after checking it's empty. Between the check and the assignment, another thread may have created a new stream.

boris-petrov:
> Hi @boris-petrov
> Thank you for PR.
> May you please rebase on `master` to pick up the latest to make sure we are still green.

Done.

> Previously, at most one `OutputStream` was allowed _per thread_. With your change, at most one `OutputStream` is allowed _globally_. This might break applications.
> 
> It's about this piece of code:
> 
> https://github.com/apache/commons-vfs/blob/a19631a852cdef62533566c158788f1bd221fa50/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/DefaultFileContent.java#L532-L534

Will check this later/tomorrow.

> Hi @boris-petrov
> 
> > This also fixes inability to close a stream from a different thread
> 
> That's one you can write a test to prove the fix actually works.

Will do.

> One more bug: TOCTOU bug in all functions which set `this.fileContentThreadData = null` after checking it's empty. Between the check and the assignment, another thread may have created a new stream.

That's pretty much the issue I explained in the original post. As I said there, it was always there and could/should be fixed by another PR.

MaxKellermann:
> That's pretty much the issue I explained in the original post. As I said there, it was always there and could/should be fixed by another PR.

No, that one was not there before, because it was manipulating a thread-local. No other thread has access to that thread-local, so there can't be a TOCTOU bug.

boris-petrov:
@MaxKellermann - you're right. I fixed this by just synchronizing all calls to `getFileContentThreadData`. They are methods that should not take long so this won't be a performance problem. And anyway they generally won't be called at the same time.

As for the at most one `OutputStream` globally - I believe that to be a feature, not a bug. Why would anyone one to write to a single file from multiple threads at the same time?

@garydgregory - I added a couple of tests.

boris-petrov:
hm, actually this means I can remove the `synchronized` from all methods in `FileContentThreadData` if I'm not mistaken. Will do that now.

MaxKellermann:
> I fixed this by just synchronizing all calls to `getFileContentThreadData`.

Be careful with that! Existing code took care *not* to run certain function calls in a `synchronized` block:
https://github.com/apache/commons-vfs/blob/master/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/DefaultFileContent.java#L91-L99
But now all callers of that method are `synchronized`, nullifying that effort.

boris-petrov:
Oh, my, there is a `openStreams` integer variable... :smile: Why isn't that used instead of `fileContentThreadData`?

boris-petrov:
And why isn't `streamClosed` called in `close` for any of the streams...?

MaxKellermann:
> Oh, my, there is an `openStreams` integer variable... smile Why isn't that used instead of `fileContentThreadData`?

Because `FileContentThreadData` is per-thread, and `openStreams` counts all streams of all threads. Your change eliminates that distinction. The real question is: why was there ever that distinction?

`FileContentThreadData` was added in 2004 by commit f299cbf1c894817bf70c96b8efd5ad80741d445b, and `openStreams` was added in 2005 by commit b1f1bbf6f576ececada5add29d6eda9cbce68eda, both by the same guy. Both commits look questionable.

MaxKellermann:
> And why isn't `streamClosed` called in `close` for any of the streams...?

It is, via `endInput()`, `endRandomAccess()`, `endOutput()`.

boris-petrov:
> > And why isn't `streamClosed` called in `close` for any of the streams...?
> 
> It is, via `endInput()`, `endRandomAccess()`, `endOutput()`.

Sorry, I meant `close` of `DefaultFileContent`. Which "manually" closes all streams but doesn't call `streamClosed` for any of them. Perhaps it's not needed because after closing a `DefaultFileContent`, you're not supposed to use it any more or something.

boris-petrov:
I pushed a commit that uses fine-grained locking. But yes, this whole thing should be revised and we should see if the `openStreams` variable isn't the only one that we need.

garydgregory:
Thank you all for your efforts.
May I ask the we look at two goals?

1. What else do we need for 2.8.1?
2. Longer term fixes.

boris-petrov:
I squashed the commits into two.

@MaxKellermann - any other suggestions (apart from the fact that we might not need `fileContentThreadData` at all and we could use `openStreams` instead?

@garydgregory - it's nice to have this fix in 2.8.1. We can remove `fileContentThreadData` and use `openStreams` - that's all that we need long-term I guess. It could happen either now (in this PR) or later. Let's see what @MaxKellermann thinks - he knows this code good obviously and has a good eye for catching mistakes.

boris-petrov:
Ah, no, this still doesn't work. In `DefaultFileContent.close` the `close` methods of the streams will call `streamClosed` in a synchronized context which is unwanted...

MaxKellermann:
Now `close()` is completely `synchronized`, which may hold the lock for a long time while the underlying I/O code is called.

MaxKellermann:
Also be careful to document what must be locked by the caller. It is dangerous that `getFileContentThreadData()` is not `synchronized` anymore, but you didn't document that synchronization is the caller's responsibility. In cases like that, I'd leave the `synchronized` in - this only adds negligible overhead, but is safe-by-default.

boris-petrov:
> Now `close()` is completely `synchronized`, which may hold the lock for a long time while the underlying I/O code is called.

Which is fine as anyway no other methods should be called on the `DefaultFileContent` instance any more. But there is a problem with what I mentioned above:

> Ah, no, this still doesn't work. In `DefaultFileContent.close` the `close` methods of the streams will call `streamClosed` in a synchronized context which is unwanted...

Which I have no idea how to fix...



> Also be careful to document what must be locked by the caller. It is dangerous that `getFileContentThreadData()` is not `synchronized` anymore, but you didn't document that synchronization is the caller's responsibility. In cases like that, I'd leave the `synchronized` in - this only adds negligible overhead, but is safe-by-default.

Yes, I was also thinking of adding a comment. I think it's better because making it `synchronized` won't fix the TOCTOU of users who call the method unsychronized.

MaxKellermann:
> I think it's better because making it `synchronized` won't fix the TOCTOU of users who call the method unsychronized.

Fixing TOCTOU bugs outside of that method (which this method cannot know about) is obviously only the caller's responsibility. But the method itself should be safe (or explicitly documented that it's unsafe).

boris-petrov:
@MaxKellermann - I pushed another commit to try to fix the issue in `DefaultFileContent.close`. I make copies of the streams lists and then close them while unsynchronized. Not sure of a better fix... any suggestions?

garydgregory:
@boris-petrov 
One test seems to consistently break on Java 15. I reran the build and it failed just the same.

boris-petrov:
@garydgregory - very strange. I think the previous build that failed was for Java 16 though, but it could have been the same test. I don't really have an idea why it might be failing... do you? I'm on Java 15 locally and the tests passed on my side. How do I run just this single test? Running `mvn test -Dtest=SftpMultiThreadWriteTests` fails with `java.lang.AssertionError: This test's read folder should not be null` - why? How do I run just this single test?

garydgregory:
@boris-petrov  Sadly, you've into one of the long standing issues with out tests: They were initially built on JUnit *3*, and later only partially ported to JUnit *4* here and there. This means that there are quirks where you can't do what you'd think you should, either from the `mvn` command line or from Eclipse... This might be an OS-related issue or related to the specific Java vendor in play...

boris-petrov:
I pushed a few times, 2 of those all tests passed, the last one blew up again on the same test. I added some logging, this is the result:

```
Copying threw an exception:
org.apache.commons.vfs2.FileSystemException: Could not copy "file:///home/runner/work/commons-vfs/commons-vfs/commons-vfs2/src/test/resources/test-data/test.zip" to "sftp://testtest@localhost:38865/write-tests/file44copy.txt".
	at org.apache.commons.vfs2.provider.AbstractFileObject.copyFrom(AbstractFileObject.java:300)
	at org.apache.commons.vfs2.provider.sftp.SftpMultiThreadWriteTests.lambda$testParallelCopyFromLocalFileSystem$0(SftpMultiThreadWriteTests.java:81)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at java.base/java.lang.Thread.run(Thread.java:832)
Caused by: java.io.IOException: Stream closed
	at java.base/java.io.BufferedInputStream.getBufIfOpen(BufferedInputStream.java:168)
	at java.base/java.io.BufferedInputStream.read1(BufferedInputStream.java:289)
	at java.base/java.io.BufferedInputStream.read(BufferedInputStream.java:343)
	at org.apache.commons.vfs2.util.MonitorInputStream.read(MonitorInputStream.java:155)
	at java.base/java.io.FilterInputStream.read(FilterInputStream.java:106)
	at org.apache.commons.vfs2.provider.DefaultFileContent.write(DefaultFileContent.java:874)
	at org.apache.commons.vfs2.provider.DefaultFileContent.write(DefaultFileContent.java:854)
	at org.apache.commons.vfs2.provider.DefaultFileContent.write(DefaultFileContent.java:827)
	at org.apache.commons.vfs2.provider.DefaultFileContent.write(DefaultFileContent.java:841)
	at org.apache.commons.vfs2.util.FileObjectUtils.writeContent(FileObjectUtils.java:184)
	at org.apache.commons.vfs2.provider.AbstractFileObject.copyFrom(AbstractFileObject.java:295)
	... 5 more
```

It doesn't seem related to the changes in this PR... you haven't seen this test fail randomly until now?

boris-petrov:
I've been running tests locally until they fail. I also see a few other tests failing:

```
ProviderWriteTests>AbstractProviderTestCase.runTest:196->testListChildren:470->assertSameFileSet:561 expected:<3> but was:<4>
```

and

```
ProviderDeleteTests>AbstractProviderTestCase.runTest:196->testDeleteAllFiles:116 expected:<2> but was:<3>
```

I also run the tests on `master` the same way and I also got a test failure:

```
ProviderWriteTests>AbstractProviderTestCase.runTest:196->testListChildren:460->assertSameFileSet:561 expected:<1> but was:<2>
```

So there is some issue with some tests... the question is if the failures that we see here are because of the PR; are because of previous problems in `master`; or are just made more often...

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/FileContentThreadData.java
```diff
@@ -22,9 +22,6 @@
 import org.apache.commons.vfs2.FileSystemException;
 import org.apache.commons.vfs2.RandomAccessContent;
 
-/**
- * Holds the data which needs to be local to the current thread
- */
```
garydgregory, line 27: -1 Uh? If anything we need more documentation, not less!

garydgregory (changes requested):
Well, we can't merge a PR that always fails the build :-( 
Maybe there is a separate PR we need to bulletproof the tests, or reveal some other bad behavior.

boris-petrov:
@MaxKellermann - do you see anything obviously wrong with this commit? Any ideas why the tests might fail (more often)?

boris-petrov:
Ah, I see now what the problem is... `getContent` on a `FileObject` is called from multiple threads (which get the same `DefaultFileContent` object) and then one of the threads closes the `DefaultFileContent` which will close all streams, including the ones for the other threads... I don't really know how to fix this easily. @MaxKellermann, @garydgregory - any ideas?

boris-petrov:
Perhaps we could not close the streams when the `DefaultFileContent` is closed but rather when the `FileObject` is closed?

MaxKellermann:
> Ah, I see now what the problem is... `getContent` on a `FileObject` is called from multiple threads (which get the same `DefaultFileContent` object) and then one of the threads closes the `DefaultFileContent` which will close all streams, including the ones for the other threads...

D'oh, so that was another reason why a ThreadLocal was used - `close()` is supposed to close only the streams by the current thread, even though the documentation says it shall close *all* streams:
https://github.com/apache/commons-vfs/blob/90e436e82831d25c9a9a8019f7099422cf6f5485/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/DefaultFileContent.java#L434
The new behavior is correct according to the API documentation, the old behavior was not correct (at least it did not do what the API documentation said it would do).

The whole idea of this `close()` method is bad. It's not a well designed API. But it's there, and must not be changed, because applications may depend on it.

> Perhaps we could not close the streams when the `DefaultFileContent` is closed but rather when the `FileObject` is closed?

That would be an API change as well. No matter how bad the API is, it must be stable. (Unless somebody decides to raise this library's major version and break the API.)

boris-petrov:
Yep, I was thinking the same things as you. So yes, there isn't much we can do without a major version in which I personally would make `FileContent` NOT extend `Closeable` - I've always wondered why it's made that way...

In any case, my original problem with not being able to close a stream from another thread is obviously unsolvable now. Unless @MaxKellermann has any other ideas, @garydgregory can close the PR and we can talk about it again for VFS 3. :)

boris-petrov:
@MaxKellermann - the only thing I would suggest is we fix the NPE at least. It's better to leak objects than to blow up. I've been checking my code... I have a million places where streams are closed in a different thread so unless the NPE is fixed, I can't use 2.8.x. And I'm pretty sure there's many people like that. What do you think?

MaxKellermann:
Since the leak bug has always been there, and given the bad existing code design, I guess adding a `null` check is the best we can do short-term. This restores the old behavior, with the old bug, but without crashing.
I don't like that, but I dislike so much about this code.
If you write a PR, maybe add a TODO comment explaining the bug and why these checks should not be necessary, to give other people an overview of what's wrong. All kludges should be well-documented to avoid confusion.

garydgregory:
Hi All:
CC: @rgoers 

Looking at git master, all the way back to to 1.0, unless I'm mistaken, in `FileContent` (for the current methods all the way back), I see seven (7) Javadocs that contain 

> There may only be a single input or output stream open for the file at any time.

- `getInputStream()`
- `getInputStream(int)`
- `getOutputStream()`
- `getOutputStream(boolean)`
- `getOutputStream(boolean, int)`
- `getOutputStream(int)`
- `getRandomAccessContent(RandomAccessMode)`

So it seems like the intent is that there should only be one input stream and one output stream in use at once.

The code does not enforce that, I'm not sure if it ever did though. 

So for now, I plan on creating a release candidate for 2.8.1 soon (maybe tonight or this weekend depending on my availability) as git master is now. I'm not sure how to move forward here. We can't break binary compatibility, so that one is simple. For behavior, I think someone needs to look at the git history to see how the code's evolved in this area and propose something, if anything.

commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/DefaultFileContentTest.java
```diff
@@ -167,4 +174,32 @@ public void testOutputStreamBufferSizeNegativeWithAppendFlag() throws Exception
...
+                check[0] = true;
+            });
+            thread.start();
+            thread.join();
+            Assert.assertTrue(check[0]);
+        }
+    }
+}
+
+interface FunctionThrowingException<T, R> {
```
garydgregory, line 203: FYI: I think you can reuse `org.apache.commons.lang3.function.FailableFunction` here.

ecki:
Imthink the Code I have Seen was supporting single output stream XOR multiple input streams. I can dig deeper for the source if you like, but I think this is also the most useful definition (if we define that formally. We do still have a gap in the area of closing a FileObject in terms of thread locks, file contents, though)

boris-petrov:
I'll close this PR until it is decided what to do with the issue. Thank you all for the time!

PR #172, opened 2021-03-28
VFS-802 - Unused return value

arturobernalg:
Unused return value.  Result of method call ignored.  Return values are never used when called


garydgregory:
Nah, we might need those.

PR #174, opened 2021-03-28
VFS-804 - Fix and Refine Javadoc

arturobernalg:


commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/ram/RamFileSystem.java
```diff
@@ -193,9 +193,9 @@ public void importTree(final File file) throws FileSystemException {
     /**
      * Import the given file with the name relative to the given root
      *
-     * @param fo
-     * @param root
-     * @throws FileSystemException
+     * @param fo the source {@link FileObject} file to import
```
garydgregory, line 196: Sentences end with a period, which did nicely elsewhere in this PR.

arturobernalg: Yep. Changed

PR #213, opened 2021-09-09
refactor CountingListener to improve test design

wx930910:
Fixes [VFS-809](https://issues.apache.org/jira/browse/VFS-809)

### Description
Replace test class [CountingListener](https://github.com/apache/commons-lang/blob/69c9593cc1da760bb4dbcf32f4ae755c54376b77/src/test/java/org/apache/commons/lang3/concurrent/AtomicSafeInitializerTest.java#L67) by mocking object created by Mockito.

<hr>

##### Key changed/added classes in this PR
- Create mocking object to replace test subclass `CountingListener`, decouple test from production code.
- Extract AtomicLong variables out of the test subclass. Use the extracted variables in assertation statement to check methods invocation status and make test condition more explict.

<hr>

### DIFF
--- a/commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/DefaultFileMonitorTest.java
+++ b/commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/DefaultFileMonitorTest.java
@@ -23,7 +23,6 @@ import java.io.BufferedWriter;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
-import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.commons.vfs2.AbstractVfsTestCase;
 import org.apache.commons.vfs2.FileChangeEvent;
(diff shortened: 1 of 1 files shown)


garydgregory:
This PR looks quite complex compared to the exting code. Why make maintenance more of a burden going forward?  What am I missing here?

wx930910:
> This PR looks quite complex compared to the exting code. Why make maintenance more of a burden going forward? What am I missing here?

@garydgregory Thanks for replying, actually we can get rid of all overridden methods as well as the attributes for tracking the overridden method exectuion by using Mockito. We can directly verify the method execution time by using `Mockito.verify()`, do you think the logic and maintenance will be better if we do it in that way?

wx930910:
@garydgregory Changed the code accordingly, no longer required AtomicLong variables and overridden methods, we just need to change the assertion to `Mockito.verify()`.

wx930910:
@garydgregory Thanks for your feedback and I changed code logic, do you think the updated code is less complex compared to the existing code? (Given the fact that we no longer need the test child class and all overridden methods)

garydgregory:
Same comments as I wrote in the other PR RE Mockito.

garydgregory:
Closing, no more feedback.

PR #222, opened 2021-11-25
VFS-810: Use different UriBuilder for Http4FileSystem

JanAelb:
There is an issue when parsing urls containing backslashes in Http4FileSystem and derived classes.

https://issues.apache.org/jira/browse/VFS-810

### DIFF
--- a/commons-vfs2-jackrabbit2/src/test/java/org/apache/commons/vfs2/provider/webdav4/test/WebDav4FilesTestCase.java
+++ b/commons-vfs2-jackrabbit2/src/test/java/org/apache/commons/vfs2/provider/webdav4/test/WebDav4FilesTestCase.java
@@ -34,7 +34,7 @@ public class WebDav4FilesTestCase extends TestCase {
      */
     @Test
     public void testUrlWithAuthority() throws FileSystemException {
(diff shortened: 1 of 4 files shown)


garydgregory:
-1 as the string in the example `"http4://alice\\1234:secret@localhost:80"` is not an RFC compliant URI, neither in [RFC 2396](https://datatracker.ietf.org/doc/html/rfc2396) or [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986).

The problem is two-fold though, the above, plus the fact that there is a long-standing issue within VFS where we incorrectly handle percent-encoded values. 

If in the above you meant to escape the a single `\` character then then URL should be `"http4://alice%5C1234:secret@localhost:80"`; but, that seems to cause problems within VFS, so that's what needs to be fixed.

JanAelb:
Thank you for the quick response. 
Passing `"http4://alice%5C1234:secret@localhost:80"` is indeed the problem I want to handle. I updated the tests.

garydgregory:
I have a possible fix in the guts of VFS, stay tuned... This PR is still not the best path IMO because it only deals with a single provider, does not solve the general problem, and adds a dependency on yet another library which happens to be used to test Hadoop.

JanAelb:
Thank you 👍

PR #228, opened 2021-12-21
[VFS-812] Don't throw FileSystemException when closing file content output stream using a BufferedOutputStream

XenoAmess:
See the test for details.

I think it strange behavior here.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/util/MonitorOutputStream.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/util/MonitorOutputStream.java
@@ -121,7 +121,9 @@ public class MonitorOutputStream extends BufferedOutputStream {
      */
     @Override
     public synchronized void flush() throws IOException {
-        assertOpen();
+        if (finished.get()) {
+            return;
+        }
         super.flush();
     }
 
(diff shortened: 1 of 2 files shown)


commons-vfs2/src/main/java/org/apache/commons/vfs2/util/MonitorOutputStream.java
```diff
@@ -121,7 +121,9 @@ public void close() throws IOException {
      */
     @Override
     public synchronized void flush() throws IOException {
-        assertOpen();
```
garydgregory, line 124: Why was the open check completely removed?

XenoAmess: I think it can be fixed this way because:

1. we will invoke **flush** when **close**.

2. after invoke **close**, every operations that will try to add something to the buffer would cause a  FileSystemException

3. then if invoke **flush** after close, there shouold actually nothing to flush

XenoAmess: So I don't think we really need to forbidden **flush** after **close**.
Just do nothing and return should be fine IMO

garydgregory:
https://issues.apache.org/jira/browse/VFS-812

garydgregory:
Applied changes based on PR #228 from XenoAmess:
- Remove cruft from unit test that was not used.
- Changed internal instance variable name from 'finished' to 'closed'
- Extract 'closed' usage into a private method.

PR #233, opened 2021-12-31
upgrade slf4j to 1.7.32

XenoAmess:
and join all slf4j version to a property, to avoid fails in https://github.com/apache/commons-vfs/pull/204 and https://github.com/apache/commons-vfs/pull/203


garydgregory:
Replaced by #244

PR #238, opened 2022-02-07
VFS-814 - FtpFileObject: re-fetch MDTM after refresh

lukewpc:
[VFS-814](https://issues.apache.org/jira/browse/VFS-814)
FtpFileObject.getTimestampMillis stores last modification time using FTPFile.setTimestamp.

doDetach() nulls out the underling FTPFile, but doesn't set mdtmSet = false this means later calls to FtpFileObject.getTimestampMillis fail to refresh mdtm and default to LIST timestamp.

---
Not sure what kind of testing you require for a patch this small?


garydgregory:
The size of the patch is irrelevant. If you fix a bug, you need a test to prove you fixed it, otherwise there nothing stopping a future change from regression to the previous behavior. The test should fail without the change to the main side of the source tree.

lukewpc:
Okay thanks, test added.

garydgregory:
Merged a while back but I want to say "thank you" here @lukewpc. Keep it up😀👍

PR #241, opened 2022-02-20
JUnit v5 upgrade reducing usage of TestCase

nhojpatrick:
Apart form pom.xml should be have merge conflicts with PR #240.


garydgregory:
@nhojpatrick 
This PR does not compile.

PR #243, opened 2022-02-21
JUnitj v5 more progression

nhojpatrick:
Everything with this pr and PR #242 now migrated to JUnit v5, unless it depends upon junit.framework.TestCase.


garydgregory:
@nhojpatrick Note that the tests are faiing.

nhojpatrick:
@garydgregory Have removed failing tests as I couldn't quickly fix, will get in the next batch.

commons-vfs2/src/test/java/org/apache/commons/vfs2/RunTest.java
```diff
@@ -23,6 +23,7 @@
 import junit.framework.TestResult;
 
 public class RunTest {
+
```
garydgregory, line 26: In the i

garydgregory (approved):
See 1 comment,  no style changes, it's annoying to review 167 files when so many contain ONLY adding or removing blank lines.

PR #257, opened 2022-04-19
VFS-818 root has read and write permissions also if he is not owner

svennissel:
If the permission of a file is 000 the user root can read and write, but not execute the file. If the permission of the file is 600 and the owner is a other user the user root can also read and write file.


garydgregory:
-1 This PR breaks the build, run "mvn pmd:check" locally for details.

garydgregory:
Rebase on git master as well please.

PR #272, opened 2022-07-12
SFTP: Memory leak because AbstractFileProvider#findFileSystem fails to detect equality of SFTP FileSystemOptions

toby1984:
When using FileSystemManager as a singleton and performing repeated SFTP operations, SftpFileSystem instances will accumulate with every resolveFile() invocation.  The cause for this is that AbstractOriginatingFileProvider#getFileSystem fails to locate the FileSystem from a previous invocation inside the TreeMap because all implementations of the SFTP IdentityProvider interface do not have equals() / hashCode() implementations and thus make calls to Arrays.deepHashcode() inside FileSystemOptions#compareTo always return different values, even if the FileSystemOption instances are actually equal.

### DIFF
--- a/commons-vfs2/src/test/java/org/apache/commons/vfs2/FileSystemOptionsTest.java
+++ b/commons-vfs2/src/test/java/org/apache/commons/vfs2/FileSystemOptionsTest.java
@@ -20,7 +20,11 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
 
+import org.apache.commons.vfs2.provider.sftp.IdentityInfo;
+import org.apache.commons.vfs2.provider.sftp.SftpFileSystemConfigBuilder;
 import org.junit.jupiter.api.Test;
+
+import java.io.File;
 
 /**
  * Check FileSystemOptions.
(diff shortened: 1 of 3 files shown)


toby1984:
Before patch:
image

toby1984:
After patch:

image

commons-vfs2/src/test/java/org/apache/commons/vfs2/FileSystemOptionsTest.java
```diff
@@ -92,4 +96,63 @@ public void testEqualsHashCodeAndCompareTo() {
...
+
+        final FileSystemOptions actual = new FileSystemOptions();
+        final IdentityInfo info2 = new IdentityInfo( privKey, pubKey, passphrase);
+        builder.setIdentityProvider(actual,info2);
+
+        assertEquals( 0, expected.compareTo( actual ) );
+        assertEquals( expected.hashCode(), actual.hashCode() );
+    }
+
+    private static void assertSftpOptionsNotEqual(File privKey1, File pubKey1, byte[] passphrase1,
```
garydgregory, line 143: Use final for parameters.

commons-vfs2/src/test/java/org/apache/commons/vfs2/FileSystemOptionsTest.java
```diff
@@ -92,4 +96,63 @@ public void testEqualsHashCodeAndCompareTo() {
...
+
+    @Test
+    public void testEqualsHashCodeAndCompareToWithSftpIdentityProviderMismatch() {
+        assertSftpOptionsNotEqual(
+            new File("/tmp/test.priv"), new File("/tmp/test.pub"), new byte[] {1,2,3},
+            new File("/tmp/test1.priv"), new File("/tmp/test.pub"), new byte[] {1,2,3}
+        );
+
+        assertSftpOptionsNotEqual(
+            new File("/tmp/test.priv"), new File("/tmp/test.pub"), new byte[] {1,2,3},
```
garydgregory, line 118: Refactor string constants that are reused into locals to make it obvious when the reuse is intended.

commons-vfs2/src/test/java/org/apache/commons/vfs2/FileSystemOptionsTest.java
```diff
@@ -92,4 +96,63 @@ public void testEqualsHashCodeAndCompareTo() {
...
+            new File("/tmp/test.priv"), new File("/tmp/test1.pub"), new byte[] {1,2,3}
+        );
+
+        assertSftpOptionsNotEqual(
```
garydgregory, line 122: IMO "Equal" -> "Equals".

garydgregory (changes requested):
@toby1984 
Thank you for your PR.
Please see my comments.

PR #277, opened 2022-07-15
Fix typos in code

marcwrobel:
Fix various typos in code :

- private field or constant names,
- method attribute names,
- test class and method names,
- exception messages.

Those changes should not be problematic : they do not belog to the public API (except maybe the exception messages).


marcwrobel:
I also found a typo in `FileSelector#traverseDescendents`, but I have created a Jira issue for it (https://issues.apache.org/jira/browse/VFS-821).

darkma773r:
Hi, @marcwrobel. Thanks for the PR! IMO, these changes fall under the "trivial" category and so don't need their own Jira issue. On the other hand, we do need the VFS-821 issue you created since that change affects the public API.

marcwrobel:
@garydgregory, you are welcome. Branch rebased against master and conflicts fixed.

PR #280, opened 2022-07-27
Handle the case where a user does not have any groups

mvdwalle:
id -g would return the gid 0
id -G however would return only a newline resulting in a NumberFormatException

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
@@ -261,19 +261,31 @@ public class SftpFileSystem extends AbstractFileSystem {
                         throw new JSchException(
                                 "Could not get the groups id of the current user (error code: " + code + ")");
                     }
-                    // Retrieve the different groups
-                    final String[] groups = output.toString().trim().split("\\s+");
 
-                    // Deal with potential empty groups
-                    this.groupsIds = Arrays.stream(groups)
-                        .map(String::trim)
-                        .filter(s -> !s.isEmpty())
-                        .mapToInt(Integer::parseInt)
-                        .toArray();
+                    this.groupsIds = parseGroupIdOutput(output);
                 }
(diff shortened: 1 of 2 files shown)


garydgregory:
Hi @mvdwalle 
This needs a unit test otherwise it might become a regression. You can refactor the code with a package private method that contains the conversion logic and call it from a unit test.

Jira: https://issues.apache.org/jira/browse/VFS-822

mvdwalle:
@garydgregory of course. Was still finding my way around the project a bit. I added a few tests to handle different output.

commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystemGroupsTests.java
```diff
@@ -0,0 +1,61 @@
...
+import com.jcraft.jsch.JSchException;
+import com.jcraft.jsch.Session;
+import org.apache.commons.vfs2.FileSystemOptions;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class SftpFileSystemGroupsTests {
+
+    FileSystemOptions options = new FileSystemOptions();
```
garydgregory, line 29: Use private and final where you can.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/sftp/SftpFileSystem.java
```diff
@@ -260,21 +261,33 @@ public int[] getGroupsIds() throws JSchException, IOException {
...
         }
         return groupsIds;
     }
 
+    /**
+     * Parses the output of the 'id -G' command
+     *
+     * @param output The output from the command
+     * @return the (numeric) group IDs.
+     * @since 2.10
```
garydgregory, line 277: You only need `since` tags in new public and private elements.

garydgregory (changes requested):
Hello @mvdwalle 
Thank you for your update. Please see my comments.

PR #322, opened 2022-10-18
VFS-825: Fix for GZip input & output stream results in StackOverflowError

japplis:


garydgregory:
@japplis 
Thank you for your PR. Merged.

PR #323, opened 2022-10-20
JUnit5 assertThrows SftpFileSystemGroupsTests

nhojpatrick:


garydgregory:
@nhojpatrick 
Uh... Is this a machine-generated PR? It does not even compile.

nhojpatrick:
> @nhojpatrick Uh... Is this a machine-generated PR? It does not even compile.

@garydgregory not all hand done, assuming it master builds on the command line i will check it builds before pushing, but might miss the odd one

PR #370, opened 2023-02-02
VFS-832: Sftp channel not put back in doGetInputStream

wangerry:
https://issues.apache.org/jira/browse/VFS-832


garydgregory:
@wangerry 
Thank you for the PR.
Would you add a unit test?

wangerry:
ok, working on it

wangerry:
> @wangerry Thank you for the PR. Would you add a unit test?

unit test is ok now.

FYI: I have to custom session facotory to test this, happy to know if you have any better idea.

i do not found equivalent sshd maxsession config in mina sshd. so i custom session factory to get channels count.

PR #383, opened 2023-03-23
fixes #VFS-834: Updated to avoid prematurely closing file objects.

ivakegg:


ivakegg:
Working on a test case that demonstrates the issue...

commons-vfs2/src/test/java/org/apache/commons/vfs2/AbstractProviderTestCase.java
```diff
@@ -114,14 +114,22 @@ protected void assertSameContent(final String expected, final FileObject file) t
      * are encoded using UTF-8.
      */
     protected void assertSameURLContent(final String expected, final URLConnection connection) throws Exception {
+        assertSameURLContent(expected, connection.getInputStream(), connection);
```
garydgregory, line 117: This does not look right, the method that allocates the resource should release it.

garydgregory:
This use case will require careful study because VFS is not built for multi-threading, and in general, getting a list of URLs should not permanently allocate resources.

ivakegg:
Gary,  This is not a multi-threading issue.  This is an issue where multiple resources are loaded for the same resource in the SAME thread.  Then the input stream of one is closed before the other is read.  Apparently this can happen with spring when loading a context.

ivakegg:
Gary,  This is not a multi-threading issue.  This is an issue where multiple resources are loaded for the same file in the SAME thread.  Then the input stream of one is closed before the other is read.  Apparently this can happen with spring when loading a context.

commons-vfs2/src/test/java/org/apache/commons/vfs2/AbstractProviderTestCase.java
ivakegg: I can add an additional try-catch however....

PR #395, opened 2023-05-07
Created a better thread safety test and fixed a couple thread safety issues

ivakegg:


commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/tar/TarFileSystem.java
```diff
@@ -16,16 +16,6 @@
  */
 package org.apache.commons.vfs2.provider.tar;
 
-import java.io.File;
```
garydgregory, line 19: Don't mess with imports, please.

commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/VfsClassLoaderTests.java
```diff
@@ -263,14 +309,14 @@ public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
...
-        executor.awaitTermination(30, TimeUnit.SECONDS);
+        executor.awaitTermination(5, TimeUnit.MINUTES);
         assertEquals(THREADS, executor.getCompletedTaskCount());
         if (!exceptions.isEmpty()) {
             StringBuilder exceptionMsg = new StringBuilder();
             StringBuilderWriter writer = new StringBuilderWriter(exceptionMsg);
             PrintWriter pWriter = new PrintWriter(writer);
             for (Throwable t : exceptions) {
-                pWriter.write(t.getMessage());
+                pWriter.write(String.valueOf(t.getMessage()));
```
garydgregory, line 319: Why?

commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/VfsClassLoaderTests.java
```diff
@@ -263,14 +309,14 @@ public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
             workQueue.addAll(rejected);
         }
         executor.shutdown();
-        executor.awaitTermination(30, TimeUnit.SECONDS);
+        executor.awaitTermination(5, TimeUnit.MINUTES);
```
garydgregory, line 312: I don't think we want to tests to possibly take "forever" when they might fail anyway.

commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/VfsClassLoaderTests.java
```diff
@@ -217,7 +249,17 @@ public void testSealing() throws Exception {
 
     @Test
     public void testThreadSafety() throws Exception {
-        final int THREADS = 40;
+        final FileSystemManager manager = getManager();
+
+        // The http4 and sftp mechanisms do not work with this thread safety test
+        List<String> schemes = Arrays.asList(manager.getSchemes());
+        if (schemes.contains("http4") || schemes.contains("sftp")) {
```
garydgregory, line 256: -1 This is not how to conditionally enable or disable tests in JUnit, and gives the _false_ impression that a tests _passes_ when it is in fact _not even run_! This kind of pattern can be handled either in each subclass in a disabled test method with a comment or using a JUnit assume statement.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/zip/ZipFileSystem.java
```diff
@@ -39,6 +29,16 @@
 import org.apache.commons.vfs2.provider.AbstractFileSystem;
 import org.apache.commons.vfs2.provider.UriParser;
 
+import java.io.File;
```
garydgregory, line 32: -1 Don't relocate import statements.

commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/ftp/FtpProviderTestCase.java
```diff
@@ -113,6 +114,14 @@ static void setUpClass(final String rootDirectory, final FileSystemFactory fileS
         if (commandFactory != null) {
             serverFactory.setCommandFactory(commandFactory);
         }
+        ConnectionConfigFactory configFactory = new ConnectionConfigFactory();
```
garydgregory, line 117: Follow the code convention in the file you are editing: In this case, use `final`.

commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/VfsClassLoaderTests.java
```diff
@@ -16,39 +16,71 @@
...
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.BrokenBarrierException;
+import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
-import org.apache.commons.io.output.StringBuilderWriter;
```
garydgregory, line 38: Don't reorder, the fewer changes in a PR, the easier it is to review.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/zip/ZipFileSystem.java
```diff
@@ -49,7 +49,53 @@ public class ZipFileSystem extends AbstractFileSystem {
...
+            @Override
+            public ZipFile initialValue() {
+                if (isPresent.get()) {
+                    throw new IllegalStateException("Creating an initial value but we already have one");
+                }
+                try {
+                    isPresent.set(Boolean.TRUE);
+                    return createZipFile(ZipFileSystem.this.file);
+                } catch (FileSystemException fse) {
+                    throw new RuntimeException(fse);
```
garydgregory, line 72: Throwing RuntimeException is an anti-pattern. Think about the problem you want to report: Is this an illegal argument, an illegal state, or something else like an actual IO exception that needs to be unchecked?

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/zip/ZipFileSystem.java
```diff
@@ -49,7 +49,53 @@ public class ZipFileSystem extends AbstractFileSystem {
...
+
+        public ZipFile getFile() throws FileSystemException {
+            try {
+                return zipFile.get();
+            } catch (RuntimeException e) {
+                if (e.getCause() instanceof FileSystemException) {
+                    throw new FileSystemException(e.getCause());
+                }
+                else {
+                    throw new RuntimeException(e);
```
garydgregory, line 85: See above.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/tar/TarFileSystem.java
```diff
@@ -54,7 +54,53 @@ public class TarFileSystem extends AbstractFileSystem {
...
+            @Override
+            public TarArchiveInputStream initialValue() {
+                if (isPresent.get()) {
+                    throw new IllegalStateException("Creating an initial value but we already have one");
+                }
+                try {
+                    isPresent.set(Boolean.TRUE);
+                    return createTarFile(TarFileSystem.this.file);
+                } catch (FileSystemException fse) {
+                    throw new RuntimeException(fse);
```
garydgregory, line 77: See my other comment about RuntimeException in another file below.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/tar/TarFileSystem.java
```diff
@@ -54,7 +54,53 @@ public class TarFileSystem extends AbstractFileSystem {
...
+
+        public TarArchiveInputStream getFile() throws FileSystemException {
+            try {
+                return tarFile.get();
+            } catch (RuntimeException e) {
+                if (e.getCause() instanceof FileSystemException) {
+                    throw new FileSystemException(e.getCause());
+                }
+                else {
+                    throw new RuntimeException(e);
```
garydgregory, line 90: See my other comment about RuntimeException in another file below.

garydgregory (changes requested):
-1 failed build and:
Quick note but the PR would need a detailed review from more folks: I am really skeptical here, VFS was _not_ designed for thread safety; hacking some code here and there is more likely to make the code base much harder to understand and maintain. In the past, we've just told users they have to handle thread safety in their applications. Slapping "synchronized" here and there is likely to create side effects like dead-locks in existing applications. While we've that've made some changes in the past to help with thread safety (see the mailing list, jira, and PRs), doing that consistently is delicate and I am concerned about unintended side effects.

commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/VfsClassLoaderTests.java
ivakegg: Because the message can be null

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/tar/TarFileSystem.java
ivakegg: sorry about that.  I understand completely.  IDE did some of this automatically.  I will undo it.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/zip/ZipFileSystem.java
ivakegg: ok, no problem.  Good point.  Thanks!

commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/VfsClassLoaderTests.java
ivakegg: ah, ok

commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/VfsClassLoaderTests.java
ivakegg: Ah right, I was doing some very long tests (many many threads) to verify things.  I will knock this back down.

commons-vfs2/src/test/java/org/apache/commons/vfs2/impl/VfsClassLoaderTests.java
ivakegg: Also it appears that the test infrastructure for the provider tests uses an older version of JUnit in which the Assume mechanism is not in place.  I guess I need to go forward and see if I can fix those two protocols to work with that test.  I would like to leave it in the provider test suite to test all of the others.  An alternative is to add a mechanism in the test suite setup in which certain test cases can be skipped.... may do that.

ivakegg:
> -1 failed build and: Quick note but the PR would need a detailed review from more folks: I am really skeptical here, VFS was _not_ designed for thread safety; hacking some code here and there is more likely to make the code base much harder to understand and maintain. In the past, we've just told users they have to handle thread safety in their applications. Slapping "synchronized" here and there is likely to create side effects like dead-locks in existing applications. While we've that've made some changes in the past to help with thread safety (see the mailing list, jira, and PRs), doing that consistently is delicate and I am concerned about unintended side effects.

The only build failure is on mac OS 8 and the error does not appear to have anything to do with my changes.  I understand the _not_ designed for thread safety issues, however if you are going to supply a class loader implementation, then the underlying protocols being used will have to be thread safe.

PR #396, opened 2023-05-16
Support URLencoding during normalization

raboof:
Earlier, the path would be URL-decoded after normalization. This meant some opportunities for normalization would be missed when dots or slashes were URL-encoded.

One way to solve this would be to change the ordering of operations, but that seems risky as it would change the meaning of some of the abstractions. Because of that I opted for a more conservative approach, where normalization will take into account URL-encoded characters, but otherwise leave the input string intact.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
@@ -47,6 +47,12 @@ public final class UriParser {
 
     private static final char LOW_MASK = 0x0F;
 
+    private static final String URLENCODED_SLASH_LC = "%2f";
+    private static final String URLENCODED_SLASH_UC = "%2F";
+
+    private UriParser() {
+    }
+
     /**
      * Encodes and appends a string to a StringBuilder.
      *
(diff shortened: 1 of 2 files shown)


commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
```diff
@@ -455,6 +455,156 @@ public static boolean fixSeparators(final StringBuilder name) {
         return changed;
     }
 
+    private static class PathNormalizer {
+        private final StringBuilder path;
+        private int cursor = 0;
```
garydgregory, line 460: Don't initialize an instance variable to its default value.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
```diff
@@ -455,6 +455,156 @@ public static boolean fixSeparators(final StringBuilder name) {
         return changed;
     }
 
+    private static class PathNormalizer {
+        private final StringBuilder path;
+        private int cursor = 0;
+        private int lastSeparator;
+        private int end;
+        PathNormalizer(StringBuilder path) {
```
garydgregory, line 463: Use final where you can.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
```diff
@@ -455,6 +455,156 @@ public static boolean fixSeparators(final StringBuilder name) {
...
+            boolean consuming = true;
+            while (consuming) {
+                consuming = consumeSeparator();
+            }
+        }
+        private void readNonSeparators() {
+            boolean reading = true;
+            while (reading) {
+                reading = readNonSeparator();
+            }
```
garydgregory, line 515: Simpler to say `while (readNonSeparator()); ?

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
```diff
@@ -455,6 +455,156 @@ public static boolean fixSeparators(final StringBuilder name) {
...
+                return false;
+            }
+            String sub = path.substring(cursor, cursor + 3);
+            if (sub.equals("%2f") || sub.equals("%2F")) {
+                cursor = cursor + 3;
+                return true;
+            }
+            return false;
+        }
+        private boolean readDot() {
```
garydgregory, line 559: Nit: Add a blank line b/w methods.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
```diff
@@ -455,6 +455,156 @@ public static boolean fixSeparators(final StringBuilder name) {
...
+            }
+            if (path.charAt(cursor) == SEPARATOR_CHAR) {
+                return false;
+            }
+            if (cursor + 2 >= end) {
+                cursor++;
+                return true;
+            }
+            String sub = path.substring(cursor + 1, cursor + 3);
+            if (sub.equals("%2f") || sub.equals("%2F")) {
```
garydgregory, line 599: Refactor magic string into a constant with a comment containing the fact that 2f is a forward slash.

garydgregory (changes requested):
Hi @raboof 
A few nits 😉

garydgregory:
@raboof 
Run `mvn` before you push to catch build errors. In this case, this PR breaks binary compatibility because it makes a previously public constructor private.

raboof:
Ah, rebase mistake... thx for the fix, not sure what's going on with `DefaultFileSystemManagerTest.testResolveFileNameType`, will look into that tomorrow.

garydgregory:
@raboof 
Please see the test failure.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
raboof: hmm, it seems PMD won't let me

garydgregory:
```
Error:  src/main/java/org/apache/commons/vfs2/provider/UriParser.java:[528,1] (whitespace) FileTabCharacter: File contains tab characters (this is the first instance).
```
Run mvn before you push to catch build errors 😉

raboof:
> Run mvn before you push to catch build errors 😉

I know ;) - some of the tests are failing or very slow on my machine (even on the main branch), probably because I'm running in a sandbox, so I've been running individual targets - sorry about the noise ;) . Now getting 500's that are not our fault it looks like though, https://www.githubstatus.com/

Do you want me to squash the commits above?

garydgregory:
Don't worry about the noise 😉 
WRT squashing, up to you, I normally squash before merging unless there's a good reason not to do so.

garydgregory:
Looks like we'll have to wait for GH to fix itself...

garydgregory:
Re-running builds now that GH is back up...

PR #409, opened 2023-08-02
VFS-841: copying Jackrabbit standalone Main class not to depend on wh…

woonsan:
…at's not supposed to be exposed by Jackrabbit2


garydgregory:
@woonsan 
Thank you for the PR
Run 'mvn' before you push to catch these build failures.

woonsan:
Hi @garydgregory ,

Yes, I think so. The Jackrabbit team believes that the Main class actually needs to be in standalone, not standalone-components and other projects such as commons-vfs2 need to just copy it. So they asked me to do this. ;-)

Kind regards,

Woonsan

garydgregory:
Hi @woonsan 
Understood. All we need is for this PR not to break the build ;-) Would you please update this PR?

garydgregory:
@woonsan 
Ping 😉

woonsan:
@garydgregory , sorry for late. I will update this by the end of this week.

garydgregory:
> @garydgregory , sorry for late. I will update this by the end of this week.

TY!

PR #421, opened 2023-09-14
VFS-843: Close the HttpConnection after consuming the entire HttpEntity.

beise:
Instead of calling `((HttpRequestBase) request).releaseConnection()` directly after sending the request, encapsulate the HttpResponse in a try-with-resource block. This way, the connection is released after the HttpEntity is fully consumed.

### DIFF
--- a/commons-vfs2-jackrabbit2/src/main/java/org/apache/commons/vfs2/provider/webdav4/Webdav4FileObject.java
+++ b/commons-vfs2-jackrabbit2/src/main/java/org/apache/commons/vfs2/provider/webdav4/Webdav4FileObject.java
@@ -95,10 +95,10 @@ public class Webdav4FileObject extends Http4FileObject<Webdav4FileSystem> {
             try {
                 final HttpVersionControl request = new HttpVersionControl(urlStr);
                 setupRequest(request);
-                try(CloseableHttpResponse res = (CloseableHttpResponse)executeRequest(request)){
-                    //close the underlying HTTP connection which is held by the response object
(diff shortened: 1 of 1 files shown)


beise:
@garydgregory 
Thanks for the hints.

Regarding the missing TestCase: 
Unfortunately I am unable to configure the started jackrabbit webdav-server to use a content-length delimited response, so it is not possible to reproduce the reported Exception on the client side: `org.apache.http.ConnectionClosedException: Premature end of Content-Length delimited message body`. It is however happening in production.

garydgregory:
Hi @beise 
Does this only happen for files of a certain size? Maybe "larger" files we do not use in tests?

garydgregory:
https://issues.apache.org/jira/browse/VFS-843

beise:
Hi @garydgregory, 

The exception happens when I list the files in a webdav-folder via `vfs.findFiles(fileSelector)`. The folder in production contains around 160 files. I have already tried to run the VFS-testcases with a folder which contains 1000 files, but I was not able to trigger the exception.

I think it has to do with the response type from the server. I guess that if the response is content-length delimited the implementation of http4 uses the `org.apache.http.impl.io.ContentLengthInputStream`, which causes the error.

With the fix from VFS-843 it runs without an error.

garydgregory:
@beise 
The code looks fine, the builds are green but you'll have to accept that this can break in the future since it's not tested :-(

garydgregory:
@beise 
Does the "1" version of the module suffer from the same issue?

PR #423, opened 2023-09-16
VFS-844: Prevent that source files urlString is twice

beise:
URIEncoded. Prevent that the urlString of a webdav-filename is twice URIEncoded when file is moved. Results in invalid filename if the filename contains spaces.


garydgregory:
https://issues.apache.org/jira/browse/VFS-844

PR #426, opened 2023-09-19
VFS-849: Close httpConnection resources properly

beise:
https://issues.apache.org/jira/browse/VFS-849


garydgregory:
@beise Please add a link to the Jira issue in the description (when there is one).

garydgregory:
Hi @beise 
TY for your PR.
How can the tests be updated to validate this change?

beise:
Hi @garydgregory 
I wrote a testcase in 'Commons VFS Jackrabbit 2' and ran it against the internal started webdav server. But I can not reproduce this behaviour. The deadlock situation only happens in my production environment or when I run some manually tests against an external webdav server started in docker (image: k3vmcd/webdav) with  SSL (HTTPS) support.
The deadlock happens when I create around 8 files via` FileObject.createFile(..)`
In my JUnit tests I tried it with 200 Files, and it went through without a deadlock.

I will try to find a reason for the different behaviour in JUnit vs. running an external webdav server.

beise:
Hi @garydgregory 
I have updated the ticket https://issues.apache.org/jira/browse/VFS-849 regarding the tests

garydgregory:
@beise 
This is no longer needed right since #428 is now in git master?

beise:
@garydgregory 
Yes this is correct. TY

PR #438, opened 2023-10-18
[VFS-524] A URI with an IPv6 address can't be parsed out correctly

fyudanov:
https://issues.apache.org/jira/browse/VFS-524

The fix is based on the one, suggested in the Jira ticket, but also includes some corner cases covered and a number of additional unit tests, also tested end-to-end.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/HostFileNameParser.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/HostFileNameParser.java
@@ -165,16 +165,24 @@ public class HostFileNameParser extends AbstractFileNameParser {
      */
     protected String extractHostName(final StringBuilder name) {
         final int maxlen = name.length();
+        final boolean isIPv6Host = isIPv6Host(name.toString());
         int pos = 0;
         for (; pos < maxlen; pos++) {
             final char ch = name.charAt(pos);
-            if (ch == '/' || ch == ';' || ch == '?' || ch == ':' || ch == '@' || ch == '&' || ch == '=' || ch == '+'
-                    || ch == '$' || ch == ',') {
+            if (isHostNameTerminatingChar(ch, isIPv6Host)) {
                 break;
             }
         }
         if (pos == 0) {
             return null;
+        }
+
+        if (isIPv6Host && pos < maxlen) {
+            if (pos == 1) {
+                return null; // Returning empty host
+            }
+
+            pos++; // Including terminating ']' into the extracted host string for IPv6 hosts
         }
 
         final String hostname = name.substring(0, pos);
@@ -291,6 +299,10 @@ public class HostFileNameParser extends AbstractFileNameParser {
         if (hostName == null) {
             throw new FileSystemException("vfs.provider/missing-hostname.error", uri);
         }
+        if (isIPv6Host(hostName) && !isHostNameTerminatingChar(hostName.charAt(hostName.length() - 1), true)) {
+            throw new FileSystemException("vfs.provider/unterminated-ipv6-hostname.error", uri);
+        }
+
         auth.hostName = hostName.toLowerCase();
 
         // Extract port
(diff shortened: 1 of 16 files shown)


commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/http5/Http5FileObject.java
```diff
@@ -94,7 +93,7 @@ protected Http5FileObject(final AbstractFileName name, final FS fileSystem,
         final FileSystemOptions fileSystemOptions = fileSystem.getFileSystemOptions();
         urlCharset = builder.getUrlCharset(fileSystemOptions);
         final String pathEncoded = ((GenericURLFileName) name).getPathQueryEncoded(getUrlCharset());
-        internalURI = URIUtils.resolve(fileSystem.getInternalBaseURI(), pathEncoded);
+        internalURI = fileSystem.getInternalBaseURI().resolve(pathEncoded);
```
fyudanov, line 96: `URIUtils.resolve()` from HttpClient5.1  has an issue of escaping host symbols that is not working properly with IPv6 hosts. HttpClient4 and native `Java URI.resolve()` don't have this problem.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/HostFileNameParser.java
```diff
@@ -357,4 +369,21 @@ public FileName parseUri(final VfsComponentContext context, final FileName base,
...
+
+    private static boolean isIPv6HostHeadingChar(final char ch) {
+        return ch == '[';
+    }
+
+    private static boolean isHostNameTerminatingChar(final char ch, final boolean isIPv6Host) {
+        if (isIPv6Host) {
+            return ch == ']';
+        }
+
```
garydgregory, line 381: Hi @fyudanov 

Thank you for this PR! 😄 

The expression below does not look right to me or the method is misnamed, but the code sure could use some comments 😉 

Below, we have the test for the last char of a host name:
```
 return ch == '/' || ch == ';' || ch == '?' || ch == ':' || ch == '@' || ch == '&' || ch == '=' || ch == '+'
                || ch == '$' || ch == ',';
```
Reading https://datatracker.ietf.org/doc/html/rfc3986#page-49:
```
host          = IP-literal / IPv4address / reg-name
...
reg-name      = *( unreserved / pct-encoded / sub-delims )
...
unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
                 / "*" / "+" / "," / ";" / "="
```
For example, shouldn't `~` be in the test. Also the method needs comments if you intend to also test for a separator char with rest of the path like `/` and `?`. Or am I missing something?

At least this PR needs more testing with host names and URI that contain all the these special characters.

garydgregory (changes requested):
Hi @fyudanov 
Please see my comment in the code.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/HostFileNameParser.java
fyudanov: Hi @garydgregory 

I agree that the char list from
```
ch == '/' || ch == ';' || ch == '?' || ch == ':' || ch == '@' || ch == '&' || ch == '=' || ch == '+'
                || ch == '$' || ch == ',';
```
looks a bit inconsistent with RFC (at first site at least), but it wasn't introduced or changed in this PR, I've just moved it to the separate method.

Not sure, if we want to revisit it in the scope of the current change, because the IPv6 host parsing algorithm doesn't actually depend on this list :)

I could add few more tests with chars like `~`, if it would be sufficient, but to me, this would cover previously existing IPv4 cases rather than the current change.

garydgregory: More tests would be great.

fyudanov: Ok, I've added few new test cases. Those that are commented out are the cases that doesn't work like I would expect according to RFC.

However, I've ensured the behavior on the tests is exactly the same as it was before this PR, so I'm not sure we want to cover missed cases here.

PR #541, opened 2024-05-28
UriParser fix and performance improvements

japplis:
I've noticed using JFR for performance profiling that `AbstractFileObject.getChildren()` was spending most of the time in `StringBuilder.substring()` done in the `UriParser.PathNormalizer` class.
See attached flamegraphs.
profiling-UriParser-crop

Looking at the code, I first notice that the substring that was using the most of time was incorrect as it was producing a String of length 2 and we were comparing it with a String of length 3.
I've added a private method `isCursorAtUrlEncodedSlash()` to perform the test for %2f using `charAt` instead of `substring` and at the same time fix the invalid comparaison.

After fixing the test in '`readNonSeparators()`', a unit test failed:

At the same time, I've replaced the other not needed `StringBuilder.substring` of `UriParser` with `StringBuilder.charAt` as profiling data (See attached flamegraph) was showing performance costs in `readDot`.
profiling-UriParser2-crop

I've also removed the `readToNextSeparator()` as it was the exact same code as `readNonSeparators()`

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
@@ -70,8 +70,8 @@ public final class UriParser {
             if (cursor + 2 >= end) {
                 return false;
             }
-            final String sub = path.substring(cursor, cursor + 3);
-            if (sub.equals("%2e") || sub.equals("%2E")) {
+            if (path.charAt(cursor) == '%' && path.charAt(cursor + 1) == '2' &&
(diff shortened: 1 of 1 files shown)


garydgregory:
Hello @japplis 
Thank you for your PR.
This bug fix and performance improvements should be performed in separate PRs to get the bug fix in (with a matching test that fails without the fix) regardless of the performance work. A performance claim should be backed up by a JMH benchmark so that a maintained can duplicate the claim. For examples of how to do this, please see Commons BCEL, Crypto, CSV, Email, IO, Lang, and Text.
TY!
Gary

japplis:
Closing this pull request as replaced by https://github.com/apache/commons-vfs/pull/543/files and upcoming pull requests.

PR #542, opened 2024-05-28
Fix for incorrect detection of %2f as separator at a few places in UriParser

japplis:
The detection of %2f in `UriParser.readNonSeparator()` is incorrect as it always compare a String of length 2 with a String of length 3.

After the fix, a unit test failed:

While adding more unit tests, I also noticed the different handling of '/' and %2f in `UriParser.normalisePath()` and fixed it.

I don't know why it shows a big diff when I only changed a few lines in `UriParser`. I guess it may have to do with the line feed but I've read many post on internet and tried many things for 1 hour without any result. If it's a problem let me know, I'll do a clean clone again and recreate the change.


garydgregory:
Hello @japplis 
You'll need to update the PR to only show what's actually changed (IOW not the whole file).

japplis:
I'm closing with pull request as I had problems to fix the line feed in the diff in this branch. I've created a new branch with a new pull request with the same change (and fix for the large diff): https://github.com/apache/commons-vfs/pull/543

PR #543, opened 2024-05-29
Fix for incorrect detection of %2f as separator at a few places in UriParser

japplis:
The detection of %2f in `UriParser.readNonSeparator()` is incorrect as it always compare a String of length 2 with a String of length 3.

After the fix, a unit test failed:

While adding more unit tests, I also noticed the different handling of '/' and %2f in `UriParser.normalisePath()` and fixed it.


commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/UriParserTest.java
```diff
@@ -99,4 +100,24 @@ public void testTypeOfNormalizedPath() {
...
+        checkNormalizedPath("./Sub Folder%2F.%2FFile.txt", "/Sub Folder/File.txt");
+    }
+
+    private void checkNormalizedPath(String path, String normalized) {
+        StringBuilder pathBuilder = new StringBuilder(path);
+        try {
+            UriParser.normalisePath(pathBuilder);
+            assertEquals(normalized, pathBuilder.toString());
+        } catch (final FileSystemException e) {
+            fail(e);
```
garydgregory, line 120: Let the exception percolate. This pattern clutters up the code IMO.

commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/UriParserTest.java
```diff
@@ -99,4 +100,24 @@ public void testTypeOfNormalizedPath() {
...
+        checkNormalizedPath("./Sub Folder/", "/Sub Folder");
+        checkNormalizedPath("./Sub Folder/../", "/");
+        checkNormalizedPath("./Sub Folder%2f..%2f", "/");
+        checkNormalizedPath("File.txt", "File.txt");
+        checkNormalizedPath("./Sub Folder/./File.txt", "/Sub Folder/File.txt");
+        checkNormalizedPath("./Sub Folder%2F.%2FFile.txt", "/Sub Folder/File.txt");
+    }
+
+    private void checkNormalizedPath(String path, String normalized) {
+        StringBuilder pathBuilder = new StringBuilder(path);
```
garydgregory, line 115: Use `final` where you can.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
```diff
@@ -89,7 +90,7 @@ private boolean readNonSeparator() {
                 cursor++;
                 return true;
             }
-            final String sub = path.substring(cursor + 1, cursor + 3);
+            final String sub = path.substring(cursor, cursor + 3);
```
garydgregory, line 93: Should we use constants instead of magic numbers here? This would help maintenance in the future IMO.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
```diff
@@ -117,7 +118,10 @@ private boolean readSeparator() {
             }
             final String sub = path.substring(cursor, cursor + 3);
             if (sub.equals(URLENCODED_SLASH_LC) || sub.equals(URLENCODED_SLASH_UC)) {
-                cursor += 3;
+                path.setCharAt(cursor, SEPARATOR_CHAR);
+                path.delete(cursor + 1, cursor + 3);
```
garydgregory, line 122: Should we use constants instead of magic numbers here? This would help maintenance in the future IMO.

garydgregory (changes requested):
Hello @japplis 
Thank you for your PR. Please see my comments.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
japplis: I've copied the `final String sub = path.substring(cursor, cursor + 3);` from the existing `readDot();` and `readSeparator();`
Not that on of my next commit, I'll remove this substring as it is a performance bottleneck. My local branch shows a 3x performance improvement.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
japplis: This pattern of `path.setCharAt` and `path.delete(cursor + 1, cursor + 3);` is already used in `canonicalizePath()` and `decode()` methods.
There is in commons-vfs a checkstyle or pmd rule for magic numbers and it starts with number 4 and above.

commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/UriParserTest.java
```diff
@@ -99,4 +100,20 @@ public void testTypeOfNormalizedPath() {
             fail(e);
         }
     }
+
+    @Test
+    public void testPathOfNormalizedPath() throws FileSystemException {
+        checkNormalizedPath("./Sub Folder/", "/Sub Folder");
```
garydgregory, line 106: I think this PR should only address ONE claim because I am not sure this is correct according to [RFC 3986 Section 3.3](https://www.rfc-editor.org/rfc/rfc3986#section-3.3) regarding trailing "/": A path "/foo/bar" has two segments and a path "/foo/bar/" has 3 segments where the last segment is empty. At least that's how I read it. I'll also [ask](https://lists.apache.org/thread/n4nrod16m7lvs36n0bcdh9w587gg8h8d) on the [dev mailing list](https://commons.apache.org/mail-lists.html), where you are welcome to participate as well.

japplis: Hi Gary, 
 Yes, this PR should address one claim. This PR doesn't change the behavior regarding the trailing "/". This is what's already happening on master. It's just that as I was writing unit tests, I was thinking to add tests regarding what's currently happen for the "/".
 Let me know what you prefer:
* Remove the unit tests regarding the trailing "/"
* Remove all the unit tests regarding "/"
* Create a new issue that normalisePath should keep the trailing "/"

raboof (approved):
I think this PR is a net improvement: it clearly improves some cases, and while I see some inconsistencies it introduces nothing that is incorrect per se.

I think it doesn't go far enough, though: looking/experimenting with the code as part of this review made me realize some further problems with the implementation.

Based on that I think we should change `UriParser#fixSeparators` to also replace encoded slashes and dots with their unencoded counterparts. This will allow the rest of this code to be simplified considerably (rolling back most of #396).

Earlier I didn't do that because I was afraid of breaking existing behavior, but having seen what further changes would be needed to make this really work, I think the simplification outweighs the risk. I gave this a quick try in #555 (changing fixSeparators but not applying all simplifications yet), and the fact that the extensive test suite still passes improves my feeling that this is the way to go.

WDYT?

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
```diff
@@ -649,8 +653,14 @@ public static FileType normalisePath(final StringBuilder path) throws FileSystem
...
-                && path.lastIndexOf("/..") != path.length() - 3
-                && path.lastIndexOf("/.") != path.length() - 2
+                && !StringUtils.endsWith(path, "/..")
+                && !StringUtils.endsWith(path, "/.")
+                && !StringUtils.endsWith(path, "%2f")
+                && !StringUtils.endsWith(path, "%2F")
+                && !StringUtils.endsWith(path, "%2f..")
+                && !StringUtils.endsWith(path, "%2F..")
+                && !StringUtils.endsWith(path, "%2f.")
+                && !StringUtils.endsWith(path, "%2F.")
```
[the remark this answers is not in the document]
japplis: Like Gary mentioned a few days ago (See below), I think this PR should only address one claim.

japplis:
Closing this PR as it has been resolved with https://github.com/apache/commons-vfs/pull/558

garydgregory:
Thank you @japplis for contributing.

PR #547, opened 2024-06-10
Improve performance of UriParser normalize

japplis:
Here is a performance improvement of `UriParser.normalise()`.
I've replaced the slow `StringBuilder.substring()` with `StringBuilder.charAt()`. I think that `StringBuilder.subSequence()` with `CharSequence.compare()` could show similar performance improvements. Let me know if you want me to test this way.

Note that I would advise to first merge https://github.com/apache/commons-vfs/pull/543/files that fixes a bug in the test of %2f. I didn't want to reintroduce the bug with this merge request so the bug is also fixed in here. This MR may have a merge conflict with https://github.com/apache/commons-vfs/pull/543/ as it touches the same part of the code. If you want I can resolve it once https://github.com/apache/commons-vfs/pull/543/ is merged.

On my machine (Desktop from 2016) I have a 3x throughput.

JMH
Benchmark                      Mode  Cnt       Score       Error  Units
UriParserBenchmark.normalise  thrpt   50  994208,463  11590,176  ops/s

After refactoring
Benchmark                      Mode  Cnt        Score       Error  Units
UriParserBenchmark.normalise  thrpt   25  2940553,554  47617,743  ops/s

JFR/JMC flamegraphs
profiling-UriParser-crop

profiling-UriParser2-crop

### DIFF
--- a/commons-vfs2/pom.xml
+++ b/commons-vfs2/pom.xml
@@ -140,6 +140,12 @@
       <artifactId>httpcore-nio</artifactId>
       <scope>test</scope>
     </dependency>
+    <!-- JMH Performance tests -->
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-core</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <properties>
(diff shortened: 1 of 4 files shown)


garydgregory:
See also #558

japplis:
Closing this PR as resolved with https://github.com/apache/commons-vfs/pull/558

PR #556, opened 2024-06-25
Add some documentation to make the unit tests easier to follow

raboof:
See also https://issues.apache.org/jira/browse/VFS-855


garydgregory:
@raboof 
Thank you for the PR.

IMO: I think we should note that this testing framework originated when we used JUnit *3* and that we want to update it to JUnit *5* which would make some of this complex and awkward test set up no longer necessary.

raboof:
suspicious that it's `macos-13, 8` each time it seems.

https://github.com/apache/mina-sshd/blob/sshd-0.8.0/sshd-core/src/main/java/org/apache/sshd/common/session/AbstractSession.java#L519

garydgregory:
Yeah, pretty suspicious 🤔

garydgregory:
Since this is only documentation, I'm going to merge it, and see if the random-ish failure goes away with the develop-Java-macos-13 combination.

PR #557, opened 2024-06-25
Update mina version used in testing

raboof:
To see if this makes testing on JDK8 MacOs more reliable.

Still old versions, just slightly less old than before.

(removed two tests that made assumptions about APIs that aren't there anymore)


commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/sftp/SftpPermissionExceptionTestCase.java
```diff
@@ -105,7 +105,7 @@ public void testGetOutputStreamException() throws Exception {
         }
 
         // try to get created channel number.
-        final int channelId = server.getActiveSessions().get(0).registerChannel(new ChannelSession());
+        final long channelId = server.getActiveSessions().get(0).getIoSession().getId();
```
raboof, line 108: (not sure if this makes any sense, to be looked at in more detail when the change seems otherwise useful)

raboof:
hmm apparently something changed on Windows

PR #555, opened 2024-06-24
WIP: Simplify UriParser

raboof:
(not for review yet, intermediate PR to test on CI)

PR #558, opened 2024-06-26
Simplify UriParser

raboof:
The reverts https://github.com/apache/commons-vfs/pull/396 and related
changes and implements the same in a simpler way by replacing the
encoded characters already in `fixSeparators`.

This approach has a slightly higher risk at breaking existing behaviour,
but a lower risk of remaining problems in this part of the codebase. All
testcases still succeed.

This PR is intended to replace https://github.com/apache/commons-vfs/pull/543 and https://github.com/apache/commons-vfs/pull/555. It includes the testcases
from https://github.com/apache/commons-vfs/pull/543, adapted to the behaviour before https://github.com/apache/commons-vfs/pull/396.

This reverts commit https://github.com/apache/commons-vfs/commit/cb45c9453669eb2570f582f491cb0e06b0e63466.
This reverts commit https://github.com/apache/commons-vfs/commit/5399c766287a06ce002beaed89fa8af5461e15d0.

This seems only slightly (<8%) slower than #547 , which might be acceptable as it's not obvious this is a bottleneck:

```
2290973,673 ±(99.9%) 55810,967 ops/s [Average]
2123414,134 ±(99.9%) 48374,989 ops/s [Average]
```

Should probably add some more tests exercising the problems that remain after the fixes from #543

### DIFF
--- a/commons-vfs2/pom.xml
+++ b/commons-vfs2/pom.xml
@@ -140,6 +140,12 @@
       <artifactId>httpcore-nio</artifactId>
       <scope>test</scope>
     </dependency>
+    <!-- JMH Performance tests -->
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-core</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <properties>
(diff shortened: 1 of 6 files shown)


commons-vfs2/pom.xml
```diff
@@ -352,6 +358,67 @@
...
+      <properties>
+        <skipTests>true</skipTests>
+        <benchmark>org.apache</benchmark>
+      </properties>
+
+      <dependencies>
+        <dependency>
+          <groupId>org.openjdk.jmh</groupId>
+          <artifactId>jmh-generator-annprocess</artifactId>
+          <version>1.37</version>
```
garydgregory, line 374: Hi @raboof 
Do you need the version here since it's already in the dependency management section in the root POM?

garydgregory:
I'd be OK to merge. Anyone else?

japplis:
I also think it's OK to merge.
Small remark StringBuilder#deleteCharAt could have been used at line 533 & 568 of UriParser.

garydgregory:
@japplis I'm more concerned about the semantics than implementation details at this point.

garydgregory:
@raboof 
Good to merge?

raboof:
Yes: I just fixed the merge conflict with 95ddecdfd55338ff3efdf8a6dda24fc10d2425d9 and switched to `StringBuilder#deleteCharAt` for the two instances @japplis mentioned.

PR #562, opened 2024-07-03
UriParser Fix for decoding files with open square bracket

japplis:
Detection of IPv6 ip addresses should ignore decoding of %nn text. It is done at the moment in the buffer length (from decode(String) instead of where the host name should be. This result in incorrect decoding of files like 'file%20[hello%20world].txt as all text after [ (and before ] if present) is not decoded.


garydgregory:
See also #558

PR #634, opened 2024-11-21
Fixed commons-vfs sandbox module not compiling due to missing imports and library

japplis:
Fix for the sub-module `commons-vfs2-sandbox` not compiling


garydgregory:
@japplis 
TY. 
Note that running `javadoc:javadoc` is still broken when running:
```
mvn -DskipTests -P include-sandbox -pl commons-vfs2-sandbox -Djapicmp.skip
```
but we're in a better state than before 👍

PR #660, opened 2025-02-04
Improve performance of encoding Uris

japplis:
If no encoding occurred for a uri (which is quite often the case), there is no need to convert the StringBuilder to a String, just return the input String.

JMH
Before
Benchmark                   Mode  Cnt        Score        Error  Units
`UriParserBenchmark.encode  thrpt   25  5814265,085 ▒ 235637,894  ops/s`

After
Benchmark                   Mode  Cnt        Score        Error  Units
`UriParserBenchmark.encode  thrpt   25  6764614,389 ▒ 213170,120  ops/s`

16.3% improvement

I've also attached a screenshot of JMC showing that a lot of time of the `UriParser.encode` is spend in the `StringBuilder.toString`

profiling-UriParser-encode

Thanks for your contribution to [Apache Commons](https://commons.apache.org/)! Your help is appreciated!

Before you push a pull request, review this list:

- [-] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible but is a best-practice.
- [x] Each commit in the pull request should have a meaningful subject line and body. Note that commits might be squashed by a maintainer on merge.


garydgregory:
@japplis 
I merged the PR. TY for your patience.
Please test your use case with git master or a `2.11.0-SNAPSHOT` build from https://repository.apache.org/content/repositories/snapshots

japplis:
Hi Gary, thank you for merging. I've tested the latest binary and everything works.

PR #663, opened 2025-02-14
[VFS-856] Fix JunctionTests

raboof:
The test would leave a file behind, causing it to fail the next time it runs (e.g. impacting 'mvn install')

- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible but is a best-practice.
- [x] Each commit in the pull request should have a meaningful subject line and body. Note that commits might be squashed by a maintainer on merge.

### DIFF
--- a/commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/test/JunctionTests.java
+++ b/commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/test/JunctionTests.java
@@ -140,7 +140,11 @@ public class JunctionTests extends AbstractProviderTestCase {
 
         final FileObject real2 = baseDir.resolveFile("weakref.txt");
         real2.createFile();
-        assertEquals("Weak Listener was abandoned", "Listener false true false", listener2.toString());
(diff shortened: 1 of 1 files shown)


commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/test/JunctionTests.java
```diff
@@ -141,6 +141,9 @@ public void testEvent() throws Exception {
         final FileObject real2 = baseDir.resolveFile("weakref.txt");
         real2.createFile();
         assertEquals("Weak Listener was abandoned", "Listener false true false", listener2.toString());
+
+        // Make sure we don't contaminate the fs for the next time the test runs
+        assertTrue(file2.delete());
```
garydgregory, line 146: Hi @raboof 
If you want this to happen 100% of the time, as opposed to not if an assertion fails, you must put it in a `finally` block.

garydgregory:
TY @raboof !

PR #680, opened 2025-04-08
VFS-859: Fix undeclared direct dependencies

ppkarwasz:
Undeclared direct dependencies that are also transitive dependencies of an **optional** direct dependency, end up being declared by Moditect using `requires <module>` instead of `requires static <module>`.

This causes JPMS runtime errors, unless the optional dependency is present on the module path.

To solve this and rationalize the dependency tree, this change:

- Adds two verification rules that check for undeclared dependencies and useless exclusions.
- Adds undeclared dependencies to the POM files.
- Remove unused declared dependencies.

### DIFF
--- a/commons-vfs2-ant/pom.xml
+++ b/commons-vfs2-ant/pom.xml
@@ -34,35 +34,22 @@
   <description>Apache Commons VFS Ant Tasks.</description>
 
   <dependencies>
+
     <dependency>
       <groupId>org.apache.ant</groupId>
       <artifactId>ant</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>commons-logging</groupId>
+      <artifactId>commons-logging</artifactId>
+    </dependency>
+
     <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-vfs2</artifactId>
     </dependency>
-    <dependency>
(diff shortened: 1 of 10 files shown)


ppkarwasz:
Related to https://github.com/moditect/moditect/issues/259

garydgregory:
Hi @ppkarwasz 

Thank you for your help :-)

Some Java 8 and 11 builds fail with `java.lang.ClassNotFoundException: javax.jcr.Binary`

garydgregory:
> > Some Java 8 and 11 builds fail with `java.lang.ClassNotFoundException: javax.jcr.Binary`
> 
> I am fighting with [`jackrabbit-standalone`](https://mvnrepository.com/artifact/org.apache.jackrabbit/jackrabbit-standalone/1.6.5), which is a shaded artifact with some strange versions of `commons-logging` and `slf4j-api`.
> 
> Should we still keep the provider for JackRabbit 1.x? The last release is almost 10 years old.

We can drop it for 3.0 IMO but I think we should do our best to keep it in 2.x since it just works. 

Maybe there are non-uber jar JR artifacts we can depend on instead?

ppkarwasz:
> Maybe there are non-uber jar JR artifacts we can depend on instead?

I'll try with `jackrabbit-webapp`

PR #695, opened 2025-06-25
Fix Javadoc for ResourceFileSystemConfigBuilder

turbanoff:
It seems there is copy-paste error in `ResourceFileSystemConfigBuilder` javadoc. For some reason it describes that it helps with `FTP`, but it's not.


garydgregory:
@turbanoff 
Ty, merged after a tweak.

PR #698, opened 2025-07-06
Vfs 860 bom

jafarre-bi:
commons-vfs2-bom has commons-vfs2-project as a parent. When you import it into a different project, you expect to have this project's dependencies managed only, but instead, you get numerous other dependencies managed from commons-vfs2-project and its parents.
This pull request detaches the BOM from its parent, which is the only way to get rid of unwanted dependency management.
Unfortunately, as the BOM is part of the reactor, other parts of the parent POM are required for the build to succeed, and we must manually copy them. The only solution for this would be to remove the BOM from the reactor itself, as building a BOM is trivial and probably allows for a much simpler POM file to be written for it. However, the BOM would then need to be built separately.

### DIFF
--- a/commons-vfs2-bom/pom.xml
+++ b/commons-vfs2-bom/pom.xml
@@ -63,5 +63,377 @@
 	<checkstyle.skip>true</checkstyle.skip>
     <!-- project.build.outputTimestamp is managed by Maven plugins, see https://maven.apache.org/guides/mini/guide-reproducible-builds.html -->
     <project.build.outputTimestamp>2025-02-14T13:36:40Z</project.build.outputTimestamp>
+    <commons.encoding>UTF-8</commons.encoding>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <!-- bare ID without major version or commons- prefix -->
+    <commons.componentid>vfs</commons.componentid>
+    <commons.packageId>vfs2</commons.packageId>
+    <commons.module.name>org.apache.commons.vfs2</commons.module.name>
+    <commons.jira.id>VFS</commons.jira.id>
+    <commons.jira.pid>12310495</commons.jira.pid>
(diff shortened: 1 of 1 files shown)


garydgregory:
The link in the description is not to the Jira issue.

So far this doesn't seem to be what Maven recommends, or am I missing something?

jafarre-bi:
> The link in the description is not to the Jira issue.

Fixed

> 
> So far this doesn't seem to be what Maven recommends, or am I missing something?

This is from [Maven documentation](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Bill_of_Materials_.28BOM.29_POMs):
_The root of the project is the BOM POM. It defines the versions of all the artifacts that will be created in the library. Other projects that wish to use the library should import this POM into the dependencyManagement section of their POM._
and
_The parent subproject has the BOM POM as its parent. It is a normal multiproject pom._

This can't be done in Apache Commons projects because they need to inherit from org.apache.commons:commons-parent.
We also can't make the BOM inherit from commons-parent because it would also inherit unwanted dependency management, namely, junit and mockito. Avoiding this implies removing the dependency management section from commons-parent, which implies adding it to the parent of every single project under it.

The solutions I can think of are:

1. Leave things the way they are. Then, the only way to fix the problem would be to remove the dependency management section from commons-vfs2-parent. You would still be inheriting unwanted dependency management from commons-parent.
2. Follow the Maven documentation recommendations. Then the BOM will inherit from commons-parent, and commons-vfs2-parent will inherit from the BOM. You would still be inheriting unwanted dependency management from commons-parent.
3. Modify the BOM so that it does not have a parent, and keep it within the reactor. This way, you don't have the option to import the BOM into the parent's dependency management section.
4. Modify the BOM so that it does not have a parent, remove it from the reactor, and import it into the parent. But then you need to build the BOM independently of (and before) the other modules.

This draft PR implements option 3, but any option is acceptable for me.

PR #754, opened 2026-03-27
Add SFTP password authentication tests for Commons VFS2

VaishKumbhar:
## What
Adds `SftpPasswordAuthTest`  — new test class that verify SFTP password authentication via `StaticUserAuthenticator` in Commons VFS2.

## Why
There is no dedicated test for SFTP password-based authentication. The existing `SftpProviderTest` relies on `SftpTestServerHelper` which accepts `username == password` and also allows `UserAuthNone`, so it does not specifically validate that `StaticUserAuthenticator` credentials are correctly propagated through the VFS2 SFTP provider.

## How
- Each test spins up an inline embedded SSHD 0.8.0 server with a strict `PasswordAuthenticator` that only accepts a specific username/password pair.
- The server uses an explicit RSA 2048-bit host key instead of the SSHD 0.8.0 default DSA key, which is incompatible with Java 17+ (disabled legacy crypto algorithms).
- Server teardown uses a daemon thread with a 5-second timeout to prevent JVM hangs caused by SSHD 0.8.0 lingering sessions after failed authentication.
- Tests verify: resolving files, resolving folders (with and without trailing slash), and that wrong credentials produce a `FileSystemException`.

## Test summary
| Test | Description |
|------|-------------|
| `testResolveFile` | Connects with correct credentials, resolves a file, asserts it exists and is readable |
| `testResolveFolder` | Connects with correct credentials, resolves a directory |
| `testResolveFolderWithTrailingSlash` | Same as above with trailing slash in URI |
| `testWrongCredentialsThrowsException` | Connects with wrong credentials, asserts `FileSystemException` is thrown |

### DIFF
--- a/commons-vfs2/pom.xml
+++ b/commons-vfs2/pom.xml
@@ -116,15 +116,15 @@
       <artifactId>log4j-core</artifactId>
       <scope>test</scope>
     </dependency>
-    <!-- Test SFTP with Apache SHHd Server (MINA) -->
+    <!-- Test SFTP with Apache SSHd Server (MINA) -->
     <dependency>
       <groupId>org.apache.sshd</groupId>
       <artifactId>sshd-core</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.bouncycastle</groupId>
-      <artifactId>bcprov-jdk16</artifactId>
+      <groupId>org.apache.sshd</groupId>
+      <artifactId>sshd-sftp</artifactId>
       <scope>test</scope>
     </dependency>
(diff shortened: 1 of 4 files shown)


garydgregory (changes requested):
Hello @VaishKumbhar

Thank you for the PR.

Please see my comments.

garydgregory:
Hello @VaishKumbhar 

Thank you for your update. Note the failures on Linux and macOS.

garydgregory (approved):
LGTM. Thank you @VaishKumbhar !

PR #771, opened 2026-07-06
Restrict reflective class loading in webdav ExceptionConverter

rootvector2:
`ExceptionConverter.generate` turns a WebDAV server's DAV error response into a `FileSystemException` and hands the response's `<exception><class>` text straight to `Class.forName(...).getConstructor(String.class).newInstance(msg)`, so a malicious or compromised server can make the client load and construct any class on its classpath (CWE-470). Found while auditing the webdav providers for untrusted-input reflection sinks. The fix loads the named class without initializing it and only constructs it when it is an `Exception` subtype, which keeps reconstruction of genuine server exception types working while refusing everything else. The jackrabbit1 (`webdav`) and jackrabbit2 (`webdav4`) providers carry the same sink, so both are fixed; the new `ExceptionConverterTest` cases fail on the current code (the crafted class name is instantiated) and pass with the change.

- [ ] Read the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) if you use Artificial Intelligence (AI).
- [ ] I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute?
- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best practice.
- [x] Each commit in the pull request should have a meaningful subject line and body. Note that a maintainer may squash commits during the merge process.

### DIFF
--- a/commons-vfs2-jackrabbit1/src/main/java/org/apache/commons/vfs2/provider/webdav/ExceptionConverter.java
+++ b/commons-vfs2-jackrabbit1/src/main/java/org/apache/commons/vfs2/provider/webdav/ExceptionConverter.java
@@ -62,14 +62,17 @@ public final class ExceptionConverter {
                         msg = DomUtil.getChildText(exc, "message", null);
                     }
                     if (DomUtil.hasChildElement(exc, "class", null)) {
-                        final Class<?> cl = Class.forName(DomUtil.getChildText(exc, "class", null));
-                        final Constructor<?> excConstr = cl.getConstructor(String.class);
-                        final Object o = excConstr.newInstance(msg);
-                        if (o instanceof FileSystemException) {
-                            return (FileSystemException) o;
-                        }
-                        if (o instanceof Exception) {
-                            return new FileSystemException(msg, (Exception) o);
+                        final Class<?> cl = Class.forName(DomUtil.getChildText(exc, "class", null), false,
+                                ExceptionConverter.class.getClassLoader());
+                        if (Exception.class.isAssignableFrom(cl)) {
+                            final Constructor<?> excConstr = cl.getConstructor(String.class);
+                            final Object o = excConstr.newInstance(msg);
+                            if (o instanceof FileSystemException) {
+                                return (FileSystemException) o;
+                            }
+                            if (o instanceof Exception) {
+                                return new FileSystemException(msg, (Exception) o);
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


garydgregory:
Thank you @rootvector2 for the PR.

Running tests... Since this is the 2nd instance of this issue pattern, you might as well scan the rest of Apache Commons for the same issue.

rootvector2:
Will do. A quick code search across the org for the same construct-from-name shape only turns up `CompressException.requireNonNull` in commons-compress as a main-tree hit, and that one takes its `Class` as a literal from internal callers rather than a name off the wire, so it's fine. I'll do a deeper sweep for `Class.forName` fed by parsed input and follow up if anything real shows up.

garydgregory:
Hm, does these complications really protect an app from anything an attacker can't work around? I mean, any `Exception` class is still loaded.

rootvector2:
the real win isn't the `Exception` gate, it's dropping the one-arg `Class.forName`. that form initializes whatever class the server names, so a hostile server can fire any class's static initializer just by naming it, before any `instanceof` check runs. `Class.forName(name, false, loader)` stops that, and the type gate keeps construction from invoking arbitrary `String` constructors.

you're right that a server can still get an `Exception` subtype loaded and constructed, so this narrows the surface rather than closing it. if the complication isn't worth it, the cleaner move is to drop the reflective instantiation entirely. the named class is the server's own exception type and rarely resolves on a client classpath, so today it mostly just falls through to `ClassNotFoundException`. happy to cut it down to building the `FileSystemException` from the message if you'd rather not carry the reflection at all.

garydgregory:
Hi @rootvector2 
Thanks for the quick feedback. I'll have to noodle on this one a bit.

rootvector2:
did the deeper sweep across the rest of the live commons components. the only other place with this same wire-name-to-instantiation shape is `commons-net`'s `DefaultFTPFileEntryParserFactory`, and it's already hardened the same way: `Class.forName(key, false, loader)` with an `isAssignableFrom` gate before construction, plus a comment noting the untrusted `SYST` reply. everything else that calls `Class.forName`/`loadClass` and instantiates is driven by app-authored config/rules, a code literal, a caller-supplied `Class`, or an `ObjectInputStream` `resolveClass` override, not a class name off the wire. so nothing else to patch under this pattern.

PR #773, opened 2026-07-14
[VFS-863] Content inside brackets in directory/file names is not decoded

nicktarallo:
In VFS 2.10, content inside brackets in the file URL is not decoded. This is because it is detected to be part of an IPv6 host, although this may not be desired behavior in all circumstances (local file paths that happen to have content in brackets are also affected, despite being unrelated to IPv6).

The fix here is to ensure that we only detect a host inside of the authority as defined by RFC 3986. We do this by detecting the authority as the content between the first instance of `"//"` and the next `"/"`, since the URL takes the format `scheme://<authority>/<path>`. Brackets `outside` of the authority hold no special meaning, so content inside of these brackets will now be decoded.

Bug was most likely introduced by [this PR]( https://github.com/apache/commons-vfs/pull/438/changes#diff-731c9fd149c7d934456163a6eb8c1afebc6be0669986e8ccf5092f164ea0cd30R166)

There appears to be another old, stale PR out for the same issue with a regex approach, but I could not find an associated ticket: https://github.com/apache/commons-vfs/pull/562

AI (Claude Code) was used to write most of the code (I reviewed all output). The fix approach and the design of the test cases were provided by me to the tool.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
@@ -163,18 +163,26 @@ public final class UriParser {
         boolean ipv6Host = false;
         boolean authorityDetected = false;
         boolean inAuthority = false;
-        int consecutiveSlashes = 0;
+        boolean slashSeenBeforeColon = false;
+        // A network-path reference starts with "//" and its authority immediately follows.
+        if (count >= 2 && buffer.charAt(index) == '/' && buffer.charAt(index + 1) == '/') {
+            authorityDetected = true;
+            inAuthority = true;
+            index += 2;
+            count -= 2;
+        }
         for (; count > 0; count--, index++) {
             final char ch = buffer.charAt(index);
             if (!authorityDetected) {
                 if (ch == '/') {
-                    consecutiveSlashes++;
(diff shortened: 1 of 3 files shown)


nicktarallo:
Hi @garydgregory, this was discussed briefly in the user mailing list, made a PR to try to address the issue.

garydgregory (changes requested):
Hello @nicktarallo 
Thank you for the PR.
Please review the comments from copilot. You might need to add a test in check the main suggestion. The test suggestion seems like a test bug.

nicktarallo:
> Hello @nicktarallo Thank you for the PR. Please review the comments from copilot. You might need to add a test in check the main suggestion. The test suggestion seems like a test bug.

Hi @garydgregory, thanks for the review. I addressed both comments:
- Authority detection now only triggers on a leading // (network-path reference) or on :// following the scheme. Added regression tests for the file:/a//[...] example from the review, a :// embedded in the path after the authority, and a URI mixing an IPv6 zone ID with brackets in the path.
-  Wrapped the Files.walk stream in try-with-resources.

Also verified all 4 new test methods fail against the pre-fix code.

garydgregory:
@nicktarallo PR merged 🚀 Thank you!

PR #774, opened 2026-07-21
Reject non-ascii hex digits in UriParser percent-decoding

rootvector2:
`UriParser.decode` and `canonicalizePath` parse the two digits of a `%XX` escape with `Character.digit(ch, 16)`, which accepts non-ascii look-alikes such as the fullwidth digits and arabic-indic digits, not just the ascii `HEXDIG` that RFC 3986 permits, so the fullwidth form of `%2e%2e%2f` decodes to `../` and slips past any ascii filter run on the raw string. Found while auditing the percent-decode path after the recent VFS-863 bracket change. The fix routes both nibbles through a small ascii-only `hexDigit` helper so a non-ascii digit raises the existing `invalid-escape-sequence` error and valid `%XX` decoding is unchanged; the new `UriParserTest` cases fail on the current code and pass with the change.

- [ ] Read the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) if you use Artificial Intelligence (AI).
- [ ] I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute?
- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best practice.
- [x] Each commit in the pull request should have a meaningful subject line and body. Note that a maintainer may squash commits during the merge process.

### DIFF
--- a/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
+++ b/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
@@ -18,6 +18,7 @@ package org.apache.commons.vfs2.provider;
 
 import java.util.Arrays;
 
+import org.apache.commons.lang3.CharUtils;
 import org.apache.commons.lang3.SystemUtils;
 import org.apache.commons.vfs2.FileName;
 import org.apache.commons.vfs2.FileSystemException;
@@ -59,10 +60,7 @@ public final class UriParser {
      * @return the value 0-15, or {@code -1} if {@code ch} is not an ASCII hexadecimal digit.
      */
     private static int hexDigit(final char ch) {
-        if (ch >= '0' && ch <= '9' || ch >= 'A' && ch <= 'F' || ch >= 'a' && ch <= 'f') {
-            return Character.digit(ch, HEX_BASE);
-        }
-        return -1;
+        return CharUtils.isHex(ch) ? Character.digit(ch, HEX_BASE) : -1;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
```diff
@@ -47,6 +47,24 @@ public final class UriParser {
...
+     * <p>
+     * Unlike {@link Character#digit(char, int)}, this only accepts the ASCII digits {@code 0-9}, {@code a-f} and
+     * {@code A-F} that RFC 3986 permits in a percent-encoding, so Unicode look-alikes such as the fullwidth digits are
+     * rejected instead of silently decoded.
+     * </p>
+     *
+     * @param ch the character to convert.
+     * @return the value 0-15, or {@code -1} if {@code ch} is not an ASCII hexadecimal digit.
+     */
+    private static int hexDigit(final char ch) {
```
garydgregory, line 62: Reuse `org.apache.commons.lang3.CharUtils.isHex(char)`

garydgregory (changes requested):
Hello @rootvector2 
Please see my comment.

commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/UriParser.java
rootvector2: Done, `hexDigit` now delegates to `CharUtils.isHex`. `UriParserTest` still passes.

rootvector2: done, `hexDigit` now delegates the check to `CharUtils.isHex` and keeps `Character.digit` just for the nibble value. `UriParserTest` still passes.

